Issue #12, opened 2020-08-14
(VIDEOCOLOR T) is not wired up for X display switching to white on black


nbriggs:
The code in xlspwin.c/lisp_Xvideocolor does not swap the foreground/background colors as it is supposed to.
Test with (IL:VIDEOCOLOR T) to switch to white on black, and (IL:VIDEOCOLOR NIL) for black on white (the default)

nbriggs:
Fixed in commit f4024ab1 -- not clear if there is something that should be done about the cursor.
IIRC, on a D-machine the cursor does not invert colors when the rest of the display is inverted.

PR #16, opened 2020-10-12
Allow changing display foreground and background colors

AbeJellinek:
Some progress on Interlisp/medley#6: no full-color display yet, but this activates previously unused `fg` and `bg` options so that the kids can have their dark mode, or sci-fi hacker mode, or whatever suits you.

Usage via `run-medley`:

- Sci-fi mode: `./run-medley -fg "#008800" -bg "#111111" -xfull35`
- Dark mode: `./run-medley -fg "#ffffff" -bg "#000000" -xfull35`

Pound sign isn't required. Just there for disambiguation.

### DIFF
--- a/src/xlspwin.c
+++ b/src/xlspwin.c
@@ -45,6 +45,8 @@ extern int Current_Hot_X, Current_Hot_Y; /* X Cursor hotspots */
 
 extern char Window_Title[255];
 extern char Icon_Title[255];
+extern unsigned long Fg_Color;
+extern unsigned long Bg_Color;
 
 extern int save_argc;
 extern char **save_argv;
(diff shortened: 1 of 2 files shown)


AbeJellinek:
See Interlisp/medley#45

AbeJellinek:
Discuss at meeting before merging

masinter:
small size bitmap fonts were hand tuned to be legible on CRT black on white. The development of outline fonts depended on some innovations making the fonts render using grey-scale.

masinter (approved):
would solve #12 #934 if merged -- when I build it 
and run ldex -fg EFEFEF -bg 202020 
the cursor stays black and doesn't get lost  -- 
 with the loss of a little bit of contrast

nbriggs:
This isn't mergeable in its current state.  If we're going to do this, we need to fix things like BlackPixel/WhitePixel which should be ForegroundPixel/BackgroundPixel (and they will default to BlackPixelOfScreen()/WhitePixelOfScreen) and we have to fix the X resource database parsing to include these as all other X11 options are done.

masinter:
I was just excited to find a Proof of Concept that solved a frustrating problem of reading my screen. I make fewer typos and get fewer headaches with vifrocolor(T) with medley fonts.

https://us06web.zoom.us/rec/share/u4c7R8g_iIim9VvVPTR8mUPmPK7tsRyBO2PlZ3Pp6TgX0jKFvwppjCsUFVMKd5U.nmzITXVzwNNh1QY6 (Passcode: +%y3G9H!)

nbriggs:
Superceded by #511 which works for both X11 and SDL.  Cursor color (for X11) was already introduced by a previous pull request.

PR #17, opened 2020-10-26
Docker compatibility: makefile and headers

AbeJellinek:
- Remove duplicate makeright
- Update some includes that don't exist on modern Linux; Clang in Docker has trouble resolving them automatically


src/inet.c
```diff
@@ -21,6 +21,9 @@ static char *id = "$Id: inet.c,v 1.3 2001/12/24 01:09:03 sybalsky Exp $ Copyrigh
 #include <sys/fcntl.h>
 #include <sys/bsdtypes.h>
 #endif /* ISC */
+#ifdef LINUX
```
nbriggs, line 24: What definition is it picking up from fcntl.h?

bin/makeright
```diff
@@ -39,6 +39,9 @@
 #	X shared libraries are needed at run time.
 #
 # Hide X shared libraries from link libraries search path.
+
```
nbriggs, line 42: Let's not separate the previous comment from the relevant line of code with this unrelated (to the comment) change.

bin/makefile-linux.x86_64-x
```diff
@@ -25,7 +25,8 @@ DISPOPTFLAGS = -O2 -g3
 FPFLAGS =
 DFLAGS = -DFSERROR -DNEW_STORAGE -DLINUX -DAIX -DOLD_CURSOR -DUSETERMIOS \
 	-DBYTESWAP -DFORKCOMM -DNOFORN -DLOGINT  $(XFLAGS) \
-	-DRELEASE=351 -D__USE_BSD -DSYSVSIGNALS -DSYSVONLY
+	-DRELEASE=351 -D__USE_BSD -DSYSVSIGNALS -DSYSVONLY \
+	-D_GNU_SOURCE
```
nbriggs, line 29: What GNU_SOURCE feature(s) are we depending on?

AbeJellinek:
```
../src/timer.c:689:55: error: use of undeclared identifier 'F_SETSIG'
  if (fcntl(ConnectionNumber(currentdsp->display_id), F_SETSIG, 0) < 0)
                                                      ^
```

It spits out this error building in a fresh Ubuntu VM without GNU_SOURCE, but no error with.

src/inet.c
AbeJellinek: Ah, you're right, this seems unnecessary. I'll remove it.

nbriggs: I think we should (perhaps not in this PR) look at the signal handling.  The _GNU_SOURCE also enables the sighold()/sigrelse() compatibility interfaces, and this should probably migrate forward to the POSIX sigaction/sigprocmask/... API

bin/makefile-linux.x86_64-x
nbriggs: OK, sounds good.  What's the right process on github to amend the commit message [Use of F_SETSIG requires _GNU_SOURCE; linux/termio.h became termios.h; etc.] (and possibly squash commits) as we merge this?

src/inet.c
AbeJellinek: I agree. We should probably have a general todo category for legacy interfaces - we definitely shouldn't be calling `getwd`, for instance (main.c:714; ufs.c:555,578,587; unistd.h:526).

bin/makefile-linux.x86_64-x
AbeJellinek: I think GitHub lets you do a squash as you merge - can't see it on my end in this repo. I would say it's best for the person running the merge to do it so that we maintain the original commit history in the archived pull request.

PR #19, opened 2020-12-10
Remove editor save files.

waywardmonkeys:
It didn't seem like these files should be in the repository.


nbriggs (approved):
Agreed.  I have now reviewed the differences between the editor save files and the current versions to ensure that we were not losing any interesting history by deleting them.

PR #20, opened 2020-12-10
Use getcwd() instead of getwd().

waywardmonkeys:
getwd() has been deprecated for a long time and getcwd() should
be used instead.


nbriggs (approved):
LGTM

PR #21, opened 2020-12-10
Remove fp.h

waywardmonkeys:
This was replaced by medleyfp.h. Perhaps this happened because
some systems have an fp.h in their system includes and the naming
collision was an issue. But nothing uses fp.h and medleyfp.h is
an updated version of the contents.


waywardmonkeys:
This is missing a change. One moment.

waywardmonkeys:
Okay, this is correct now.

nbriggs (approved):
LGTM.  The shift from fp.h to medleyfp.h probably happened when support was added for OSF1, since medleyfp.h includes (the system) <fp.h> on that OS.

PR #22, opened 2020-12-11
Add initial CI via GitHub Actions.

waywardmonkeys:
This does a basic build for macOS Catalina and Ubuntu Bionic.


nbriggs (approved):
Cool.  Too bad they don't offer a [Free|Open]BSD OS option for the runners.

waywardmonkeys:
Agree. But one can run self-hosted builders as well.

But at least getting coverage for some targets is a good first step!

PR #23, opened 2020-12-11
Remove inc/lnk-*.h

waywardmonkeys:
These headers aren't used and are all old copies of the same
headers without `lnk-` in the name.


nbriggs (approved):
LGTM.  I've reviewed the differences between lnk-x and x for all x, and we're not losing any valuable history by removing these files.

PR #24, opened 2020-12-11
Remove bin/optck and bin/optck-.

waywardmonkeys:
These were binaries built for SPARC to test some assembler
optimizations.

They aren't needed as part of the version control as they're just
artifacts of the past.


nbriggs:
They're artifacts I'd like to have a closer look at (I intermittently have access to SPARC systems) before I merge this removal.

nbriggs:
OK, inspection done.  I'll merge this PR.

PR #27, opened 2020-12-11
initkbd: Fix warning about missing decl for strcmp.

waywardmonkeys:
Instead of adding more stuff to the `#ifdef` around the inclusion
of `<string.h>`, we remove that `#ifdef` entirely.

There are many instances of `<string.h>` being included in the
code without any `#ifdef` protection, and `strcmp` is defined to
be declared in `<string.h>`, so all platforms from the last 20
years and into the future will need this.


waywardmonkeys:
If you agree with this, then I might well submit a patch that fully unprotects inclusions of `<string.h>` throughout the code.

nbriggs (approved):
LGTM.

PR #28, opened 2020-12-11
Remove support for Ultrix.

waywardmonkeys:
Ultrix hasn't seen an update since 1995 and the DEC3100 hardware
that this was originally written for was released in 1989.

It isn't likely that anyone has Ultrix running on hardware where
they would also want to run this.


waywardmonkeys:
If this is acceptable, then I would like to follow up with similar targeted pull requests to remove Apollo and some other older platforms.

waywardmonkeys:
Also, I didn't remove `DEC3100` stuff because that's also used in the `OSF1` build environment.

nbriggs (approved):
Looking at the details, I'm OK with removing the Ultrix support -- there's nothing here that isn't system header or OS-specific system calls that don't contribute context to the Lisp emulation.

waywardmonkeys:
I've rebased this forward after the Apollo removal landed.

nbriggs:
Oops.  Thanks.  I got things out of order.

PR #29, opened 2020-12-12
Remove unused _I386 -D flag.

waywardmonkeys:
This flag isn't actually used in any code, so let's remove it to
avoid confusion over whether it does do something or not.


nbriggs (approved):
LGTM.

PR #30, opened 2020-12-13
Remove platform support for Apollo/Domain.

waywardmonkeys:


nbriggs (commented):
You're not going after the
```
# This is to make the %$#@! Apollo cc happy
OEXT = .o
```
that's scattered through almost all the bin/make* files?

waywardmonkeys:
I wanted to do that separately!

waywardmonkeys:
But I could do it as a second commit in this PR. It was confusing as the Apollo code was doing the same as everyone else here.

nbriggs:
A separate PR would be fine too, along with thinking about whether we actually need any OEXT -- aren't all platforms using .o and not parameterizing it at all?

nbriggs:
I'll go ahead and merge this as is.

PR #31, opened 2020-12-13
uutils: Clean up string.h inclusion.

waywardmonkeys:
This is similar to the change in #27. We can just include <string.h>
rather than doing performative dances and complicating everything.


nbriggs (approved):
LGTM.  If we ever go to try to compile the code on a platform that requires <strings.h> rather than <string.h> we'll just have to readdress it ;-)

PR #33, opened 2020-12-13
Remove SGI Irix platform support.

waywardmonkeys:
This platform had no emulation-related changes.

Remove code for IRIX40, INDIGO definitions.

In disk code, simplify a chain of #ifdef's around fsync handling.


nbriggs (approved):
LGTM.

PR #34, opened 2020-12-13
Remove HP platform support.

waywardmonkeys:
This removes code related to HP9000, HPTIMERBUG, and HPUX defines.

It leaves KB_HP9000 for now as I'm not sure about renumbering
those constants.

This should not impact any of the core emulation code.


waywardmonkeys:
Getting rid of this one is getting us closer to killing off some of the older code for ioctls and old signals code. (ISC support needs to go and we'll be even closer.)

PR #36, opened 2020-12-13
Remove OSF1 / DEC3100 platform support.

waywardmonkeys:
This also removes support for the DEC3100 keyboard, but leaves a
constant behind for it as I'm not sure about re-numbering.


src/initkbd.c
```diff
@@ -450,8 +414,8 @@ char *getenv(); /*  ---- external entry points --------*/
 #define MIN_KEYTYPE 3
 #define KB_AS3000J (7 + MIN_KEYTYPE)
 #define KB_RS6000 (8 + MIN_KEYTYPE)
-#define KB_DEC3100 (9 + MIN_KEYTYPE)
-#define KB_HP9000 (10 + MIN_KEYTYPE)  // TODO: Can we remove this?
+#define KB_DEC3100 (9 + MIN_KEYTYPE) /* TODO: Can we remove this? */
```
nbriggs, line 417: There are 3 bits allocated for the keyboard type in the Interface page devconfig word.  The system image contains that information.  Keyactions need to be reassigned when an existing system image is restarted with a different keyboard type.  Keyboard numbers can be removed, but the numbering should not be changed -- it would confuse the startup of preexisting system images.

PR #38, opened 2020-12-14
Remove FUJI compile flag.

waywardmonkeys:
This was only used for a keyboard warning that doesn't seem
relevant any longer.


nbriggs (approved):
LGTM.  We no longer have to care what Fuji Xerox thought the keyboard detection behavior should be.

PR #40, opened 2020-12-14
Begin to remove ISC platform support.

waywardmonkeys:
We don't want to modify emulation code yet as ISC support included
a number of other things, including some i386 code that isn't used
on other platforms, but is still useful as a reference.


waywardmonkeys:
This one is trickier than the others.

bin/compile-flags
```diff
@@ -12,11 +12,9 @@ _________	________________________________________________________
 BYTESWAP	Used when compiling for a hardware architecture that has
 		byte-swapped words and word-swapped 32-bit cells, e.g. 80386.
 SYSVSIGNALS	True when compiling on a system that requires the use of
-		SYSV (rather than BSD) signal-handling code (just MIPS & ISC
-		for 486, now).
+		SYSV (rather than BSD) signal-handling code.
 SYSVONLY	True when compiling on a system that has little or no
-		BSD support (the ISC unix for 486, or MIPS).  E.g. gettimofday
-		is missing.
+		BSD support.  E.g. gettimofday is missing.
```
nbriggs, line 17: s/gettimofday/gettimeofday/

waywardmonkeys: Fixed and force pushed.

waywardmonkeys:
ISC was the only platform to define `RESWAPPEDDCODESTREAM` ... but I didn't remove the code for that. Not really sure what it was about.

nbriggs:
There's a bit in the function header that says that the opcodes in the function byte code stream are in native order or not (which really means are they little-endian).
bin/ChangeLog --
```
1 Jan 02  JDS

        FILES CHANGED:  src/byteswap.c bin/makefile-tail

        Fix MAKE rules for tstsout, setsout so they use the byteswapper if
        necessary.  Changed byteswap.c to only define byte_swap_code_page
        if we're using reswapped code streams (which we aren't, as of
        now), so we don't have to drag that all in or move the function to
        someplace unrelated.
```

waywardmonkeys:
I think this is ready to land now and follow ups will happen that clean up additional things.

I have a couple of changes that I am doing that are blocked on this one.

nbriggs (approved):
LGTM.  Odd finding an #ifdef sun in the middle of a piece of ISC code, but probably historical from before the sun+other code was replaced and this was left to handle the ISC case.

waywardmonkeys:
I wonder if it has anything to do with the ISC assets being bought by Sun and, as you said, some shared code paths from perhaps the SunOS3 days. I had forgotten all about Kodak’s Unix work (which became ISC...)

waywardmonkeys:
Err, ISC went to Kodak and then Sun, not originating at Kodak.

PR #42, opened 2020-12-14
Remove 1988-era copies of some SunOS headers.

waywardmonkeys:


nbriggs (approved):
LGTM.  The dependencies are a complete mess, but I suspect it will take reworking the build infrastructure before we can get it right.  ATM I "makeright x cleanup && makeright x" rather than depend on the listed dependencies.

PR #43, opened 2020-12-14
Use tcgetattr/tcsetattr in termios code, not ioctl.

waywardmonkeys:
Previously, we were using ioctl directly on some platforms rather
than always going through termios. This complicates portability.


waywardmonkeys:
This change would make @masinter 's life easier with cygwin as it won't need to modify the code now to go through the correct code path.

nbriggs (approved):
LGTM

PR #44, opened 2020-12-14
Use waitpid, not wait3, etc.

waywardmonkeys:
This removes code that uses `union wait` status in favor of code
that uses `int` status. This has been the correct way since
SvR4, BSD 4.3, POSIX.


nbriggs (approved):
LGTM

PR #46, opened 2020-12-14
Remove AIX RS/6000 platform support.

waywardmonkeys:
AIX on PS/2 will be removed separately.


src/initkbd.c
```diff
@@ -627,11 +602,6 @@ void keyboardtype(int fd)
       SUNLispKeyMap = SUNLispKeyMap_for4;
       InterfacePage->devconfig |= type - MIN_KEYTYPE; /* 7 */
       break;
-    case KB_RS6000:
```
nbriggs, line 630: This is not a blocking issue for this change, but we need to document the values that went into the<code>InterfacePage->devconfig</code> for all the keyboards that have been supported.

nbriggs (approved):
LGTM

PR #48, opened 2020-12-15
Remove src/xc.c.orig

waywardmonkeys:
This was a backup copy of xc.c made in 2001 (judging by the Id
line).


nbriggs (approved):
"xc" is the main dispatch loop of the emulator, so any non-trivial differences here are important.   I have prettyprinted both xc.c and xc.c.orig and compared them, and there are no significant differences.  So, LGTM.

PR #49, opened 2020-12-15
Update an inclusion of inlnPS2.h

waywardmonkeys:
Most includes of this file are protected by #ifdef GCC386, but this
one was AIXPS2, which is going away.


nbriggs (approved):
I agree with this.  The inlnPS2.h file provides inline assembler intended for 32-bit x86 systems where the code is compiled with GCC circa 2002.

While there are currently no systems defining GCC386, that's because it was easier to port with pure C code rather than deal with the inline assembler, and it is not necessarily abandoned.

PR #50, opened 2020-12-15
Remove AIX on PS/2 platform support.

waywardmonkeys:


src/dsk.c
```diff
@@ -2377,19 +2372,14 @@ LispPTR COM_getfreeblock(register LispPTR *args)
...
-#endif /* AIXPS2 */
-
   if (rval != 0) {
 #endif
     *Lisp_errno = errno;
     return (NIL);
   }
 #if defined(SYSVONLY) || defined(OS5)
   *buf = sfsbuf.f_bfree;
-#elif (!defined(AIXPS2))
```
nbriggs, line 2391: This can't be correct -- at least MACOS and FREEBSD would fall into this case, and really the OS5 case should be using f_bavail (blocks available to non privileged user) rather than f_bfree (free space on file system).  I don't know what fields were in the statfs struct for SYSV.

waywardmonkeys: Yes, I think I made an error here as well.

The currently supported Unix OSes (in theory) at this point are FreeBSD, Linux, macOS, and Solaris (OS5).

So I think I can rewrite this a bit better here and then I'll force push.

Thanks for catching this!

waywardmonkeys: So, since AIXPS2 was never defined, we were never hitting the 200000 fallback path and always hitting `f_bavail`. I've fixed this here now.

After this is in, i'll submit a follow up PR that makes us use the POSIX `statvfs` and stops using the non-portable `statfs`.

nbriggs: FreeBSD <code>struct statvfs</code> does not include the f_bavail field, they note it's a wrapper around <code>statfs</code> and does not include all fields available through that syscall.

waywardmonkeys: MacOS claims the same thing in its man page, but the actual struct does. I'll check FreeBSD further. (The POSIX docs for `statvfs` say that it includes `f_bavail`.)

Sad that in 2020 things are still so broken!

nbriggs: Ah, you're right -- both macOS and FreeBSD include it though fail to state that in the man page.  Sad indeed!
```
struct statvfs {
        fsblkcnt_t      f_bavail;       /* Number of blocks */
        fsblkcnt_t      f_bfree;
```

nbriggs (approved):
Looks good now.

src/dsk.c
waywardmonkeys: Oddly enough... OpenBSD gets this right (since 2008?) and FreeBSD apparently still gets this wrong-ish. macOS seems to get it right.

PR #51, opened 2020-12-15
Remove the signal handler prototype that returns int.

waywardmonkeys:
These should return void and we no longer build for any OS where
an int is returned.


waywardmonkeys:
I will fix the arguments in a subsequent PR.

This is another that should help out @masinter with Cygwin.

nbriggs (approved):
Agreed.  I think we may be able to shift to <code>sigaction</code> without too much difficulty at this point.

PR #52, opened 2020-12-15
Modernize / simplify statfs / statvfs handling.

waywardmonkeys:
On FreeBSD, we still need to use `statfs`, but everywhere else (in
Unix land), we should be able to use `statvfs`.

We have to use `statfs` on FreeBSD as the `statvfs` there is not
like on other platforms and doesn't expose `f_bavail`, while
`statfs` does.


waywardmonkeys:
I need to revise this given that FreeBSD does actually have the `f_bavail` member. I'd missed it when I looked at the source somehow.

waywardmonkeys:
Okay, this should be ready now. So much cleaner!

nbriggs (approved):
Looks much better.

PR #54, opened 2020-12-15
Remove remnants of valloc workarounds

waywardmonkeys:


waywardmonkeys:
valloc is page-aligned but isn't commonly used these days. But maybe we'd like to align this allocation in some other way? Or it isn't important?

nbriggs:
Not sure about this one.   Every Unix system we compile for seems to support valloc(), and it makes debugging easier when the Lisp pages are aligned with the system pages.

waywardmonkeys:
Okay, I'll modify this PR then ... one moment.

nbriggs:
We could go with posix_memalign() if it's support on all the systems we support (and I think that now, it is)

waywardmonkeys:
Looks like `valloc` is now more standard than it used to be, so instead of removing it, I have now removed the workarounds for it not being everywhere.

PR #56, opened 2020-12-15
Remove unused BSD_COMP define from SunOS5 config.

waywardmonkeys:


nbriggs:
The code in rpc.c will not compile without defining BSD_COMP -- it controls the definitions of constants in the Solaris ioctl.h.  Without BSD_COMP, it fails with
```
"../src/rpc.c", line 128: undefined symbol: FIONBIO
```

waywardmonkeys:
Ahh, I see. That code should be fixed to use `fcntl` with `O_NONBLOCK` ...

PR #58, opened 2020-12-15
Improve I_SETSIG check for X I/O.

waywardmonkeys:
I_SETSIG is part of the STREAMS interface, which is only present
on Solaris (among our supported platforms), so simplify this check.

Other usages of I_SETSIG could be improved, but they need more
investigation and perhaps removal.


nbriggs (approved):
LGTM.

PR #60, opened 2020-12-15
Convert nearly-dead code to using statvfs, not statfs. Remove unused …

waywardmonkeys:
…stat buf.

This code is for non-DOS, non-XWindows, and still used the legacy
statfs API.


nbriggs (approved):
LGTM.

PR #63, opened 2020-12-15
Use LINUX, not AIX, in medleyfp.h.

waywardmonkeys:
AIX was defined by a number of targets, but the only one that
doesn't currently have anything here is LINUX, so we can just
check that instead.

### DIFF
--- a/inc/medleyfp.h
+++ b/inc/medleyfp.h
@@ -60,7 +60,7 @@ volatile extern int  FP_error;
 #define FPCLEAR
 #define FPTEST(result) (isinf(result) || isnan(result))
 
-#elif defined(AIX)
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
The actual change LGTM.
The system includes and definitions of FPTEST need a separate review.  Solaris2 has at least three sets of floating-point functions and includes [of differing levels of deprecation] (ieeefp.h, math.h, sunmath.h), and finite() became isfinite() in c99 (according to Solaris).  I haven't reviewed what Linux, FreeBSD, and macOS do.

waywardmonkeys:
@nbriggs I filed issue https://github.com/Interlisp/medley/issues/79 about that very subject earlier!

PR #64, opened 2020-12-15
Switch from ioctl/FIONBIO to fcntl/O_NONBLOCK.

waywardmonkeys:


waywardmonkeys:
This is only built on Solaris, so if you could give it a try there, I'd appreciate it!

nbriggs (approved):
Compiles OK on Solaris 11.4/sparc.  I don't have an RPC based test case to check operation.

waywardmonkeys:
@nbriggs We can probably bring back that `BSD_COMP` removal patch then!

nbriggs:
@waywardmonkeys Yes we can -- I've just verified that it compiles cleanly without it.

PR #65, opened 2020-12-15
Use settimeofday() instead of stime().

waywardmonkeys:


src/timer.c
```diff
@@ -335,15 +335,11 @@ void subr_settime(LispPTR args[])
   dosday.year = uxtime.tm_year;
   dosday.dayofweek = uxtime.tm_wday;
   _dos_setdate(&dosday);
-
-#elif defined(SYSVONLY) && !defined(MACOSX) && !defined(LINUX)
-  time_t newTime = (time_t)(*((int *)Addr68k_from_LADDR(args[0])) - UNIX_ALTO_TIME_DIFF);
-  stime(&newTime);
 #else
   struct timeval timev;
   timev.tv_sec = *((int *)Addr68k_from_LADDR(args[0])) - UNIX_ALTO_TIME_DIFF;
```
nbriggs, line 340: Bug fix: <code>timev.tv_usec = 0;</code>

waywardmonkeys: Good catch.

PR #66, opened 2020-12-15
Remove more AIX (but not really AIX) ifdefs.

waywardmonkeys:
Trying to remove / replace / update all remaining AIX ifdefs.

### DIFF
--- a/src/bbtsub.c
+++ b/src/bbtsub.c
@@ -81,13 +81,6 @@ static char *id = "$Id: bbtsub.c,v 1.3 2001/12/24 01:08:59 sybalsky Exp $ Copyri
 
 extern IOPAGE *IOPage68K;
 
-#ifdef AIX
-#include <sys/types.h>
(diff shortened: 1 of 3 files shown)


src/bbtsub.c
```diff
@@ -81,13 +81,6 @@ static char *id = "$Id: bbtsub.c,v 1.3 2001/12/24 01:08:59 sybalsky Exp $ Copyri
 
 extern IOPAGE *IOPage68K;
 
-#ifdef AIX
-#include <sys/types.h>
-#if !defined(LINUX) && !defined(MACOSX) /* LINUX and MACOSX have no macros.h */
-#include <macros.h>
```
nbriggs, line 87: Solaris has <macros.h> for min, max, etc. but I think perhaps we don't want to include it just for that since it also defines global variables.

nbriggs (approved):
LGTM

PR #72, opened 2020-12-15
Update darwin x11 location

nbriggs:
XQuartz installations on macOS don't install a symbolic link from /usr/X11 to /opt/X11 (because system integrity protection is enabled?) so we should use /opt/X11/lib directly as the location of the X libraries, as we had already done for the location of the include files.


nbriggs:
oops, included the prior PR changes.

PR #74, opened 2020-12-16
Remove final instances of index/rindex defines.

waywardmonkeys:


nbriggs:
Yeah, I hadn't removed those because it was a file from NFSWATCH, but I guess it doesn't really matter.

PR #76, opened 2020-12-16
Fixup xdefs.h

nbriggs:
Update guard name to match file name and protect entire file
Only declare extern lock variables when doing locking
Include system headers that locking macros depend on
Convert code block  macros to do { } while (0) statement style


waywardmonkeys:
LGTM from visual review.

PR #77, opened 2020-12-16
unixfork: read() returns ssize_t.

waywardmonkeys:
We should do this more widely, but this will let us see if we
have any problems across platforms from making this change.


nbriggs (approved):
LGTM.
This should be fine here, but we need to be careful when converting back to a Lisp number.

waywardmonkeys:
@nbriggs By allowing this, you don't know what you've unleashed. Hahahahahaha.

nbriggs:
Let's deal with issue #82 before going crazy!

PR #78, opened 2020-12-16
FD_ISSET requires that the fd being tested is >= 0

nbriggs:
test the fd is >= 0 before checking it with FD_ISSET().

### DIFF
--- a/src/initdsp.c
+++ b/src/initdsp.c
@@ -108,8 +108,8 @@ int oldred[2], oldgreen[2], oldblue[2];
 #endif /* DISPLAYBUFFER */
 #endif /* NOPIXRECT */
 
-int LispWindowFd;
-int FrameBufferFd;
+int LispWindowFd = -1;
+int FrameBufferFd = -1;
 
 int displaywidth, displayheight, DisplayRasterWidth, DisplayType;
 int DisplayByteSize;
--- a/src/initkbd.c
+++ b/src/initkbd.c
@@ -123,7 +123,7 @@ extern struct screen LispScreen;
 #endif /* SUNDISPLAY */
 
 extern int LispWindowFd;
-int LispKbdFd;
+int LispKbdFd = -1;
 extern int errno;
 
 /*   for debug    */
--- a/src/keyevent.c
+++ b/src/keyevent.c
@@ -124,11 +124,11 @@ extern fd_set LispIOFds;
 fd_set LispReadFds;
 int XNeedSignal = 0; /* T if an X interrupt happened while XLOCK asserted */
 
+extern int LogFileFd;
+
 #ifdef NOETHER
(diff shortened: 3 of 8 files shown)


waywardmonkeys:
Are these values getting initialized to `0` due to being in the BSS? Are they then being used prior to being initialized?

I think that `FD_ISSET` is supposed to support `0 <= fd < FD_SETSIZE`?

nbriggs:
They're initialized to -1, and set when the particular entity is opened.

I hit a SEGV in the debugger for FD_ISSET(-1, xxx) for this, which I am fixing before I fix the real problem... which I believe is actually a SIGPOLL on the X connection, where we've enabled SIGPOLLs, before we have the signal handler set up... so it exits.

waywardmonkeys:
`LogFileFd` is never initialized outside of the place where it is set to `cons_pty`. Interestingly, it is also a `u_int` while others are `int`.

The `ether_fd` looks good, and is initialized at the point of declaration.

The `RS232C_Fd` is initialized in an init function. It would be nice if it were initialized like the `ether_fd`.

The `FrameBufferFd` isn't discussed here, but is similar to the `RS232C_Fd`.

The `LispWindowFd` is also only initialized in the init function.

Is it guaranteed that nothing can access these values before the init functions are called?

nbriggs:
The SEGV was because the ether_fd was -1 when the X display connection generated a SIGPOLL event.

On Dec 15, 2020, at 9:40 PM, Bruce Mitchener <notifications@github.com> wrote:
> 
> LogFileFd is never initialized outside of the place where it is set to cons_pty. Interestingly, it is also a u_int while others are int.

It should be an int, the fact that LogFileFd exists at all is sketchy, it could just use cons_pty.
> The ether_fd looks good, and is initialized at the point of declaration.
> 
> The RS232C_Fd is initialized in an init function. It would be nice if it were initialized like the ether_fd.
> 
> 

Yes, we should be able to do that.
> The FrameBufferFd isn't discussed here, but is similar to the RS232C_Fd.
> 
> The LispWindowFd is also only initialized in the init function.
> 
> Is it guaranteed that nothing can access these values before the init functions are called?
> 
> 

They can be accessed after make_dsp_instance() is called (from the signal handler), and I suspect that the call to make_dsp_instance() can move to after the init_* functions are clled, right near start_lisp() call.

nbriggs:
I'll update this branch with a better fix (tomorrow).  All the fds can be statically initialized to -1 so there's no question of accessing uninitialized values in the signal handler.  I'll push a separate fix for the (too) early enabling of SIGPOLL/SIGIO for X window interactions.

waywardmonkeys:
LGTM, but I might have split the removal of the fcntl into a separate commit, in case something had to be reverted, the other beneficial changes would remain.

PR #79, opened 2020-12-16
main.c has no dependency upon unixfork.h.

waywardmonkeys:
This avoids a bit of confusion, since unixfork.c isn't built into
the same executable as main.c. (It is part of `lde` instead.)


nbriggs (approved):
LGTM.

PR #80, opened 2020-12-16
Update old makefiles for renaming of byteswapfns.c

waywardmonkeys:
This was renamed long ago for the DOS port, but not all of
the makefiles were updated.


nbriggs (approved):
Looks as though it should be ok.

PR #84, opened 2020-12-16
Remove commented out directives for SunOS4 dispatch loop hacks.

waywardmonkeys:
On SunOS4, we build `xc.c` by jumping through a lot of hoops to
perform a particular set of operations. We don't do that on other
platforms and this will help clean up the Linux and FreeBSD makefiles.

These directives were already not present in the macOS makefiles.


nbriggs (approved):
LGTM.

PR #85, opened 2020-12-16
Remove asmbitblt.c, duplicate of asmbbt.c.

waywardmonkeys:
When things were renamed for DOS, `asmbitblt.c` became `asmbbt.c`,
but the old file got left around. It was never converted fully
from K&R prototypes, but apart from that, is the same as `asmbbt.c`.


nbriggs:
Did you happen to notice the error in converting from K&R style in asmbbt.c?

waywardmonkeys:
The 2nd parameter should be a pointer ... but I didn't get to looking into that further.

nbriggs:
Yes, it looks as though the conversion was based on
```
    DLword *srcbase,
    dstbase;
```
missing that the pointer-ness carries over.
Since we're removing the history that gives us that hint, let's at the same time fix the missing *.
Chances are we won't be looking at hand optimizing the bitblt code, but it used to be extremely fast due to the hardware support for it on the D machines.

waywardmonkeys:
Submitted the fix for the signature as PR #89.

PR #86, opened 2020-12-16
SunOS5/SPARC and init.sparc now define FSERROR.

waywardmonkeys:
All other targets that we build were defining `FSERROR` except
these two.


nbriggs (approved):
LGTM.

PR #87, opened 2020-12-16
Remove USE_UTIME.

waywardmonkeys:
This wasn't enabled and the corresponding code wouldn't have
compiled. The `utime()` function is also deprecated in POSIX.

This used to be enabled for HPUX and RISCOS.


nbriggs (approved):
LGTM

PR #89, opened 2020-12-16
asmbbt: Fix signature for bitblt.

waywardmonkeys:
This was an error in the conversion from K&R C prototypes.


waywardmonkeys:
This was found when dealing with PR #85.

nbriggs (approved):
LGTM.

nbriggs:
I wonder how many other places have the same mistake.

PR #90, opened 2020-12-16
Remove CLX support.

waywardmonkeys:
CLX is a Common Lisp implementation of the X client library.

It has some code in C, which was present here as `src/socket.c`
and `src/socdvr.c`, exposed via opcodes in `src/subr.c`. This code
had been removed (with prejudice apparently) by commenting out
the code in `src/subr.c` with `#if NEVER`.

This code would've been used by the Medley system images, but
that code doesn't appear to be present. (There may or may not
be something related in `XMAS` in the Medley repository.)


nbriggs:
I'd like to see an opinion on this from at least @Anzus

Actually, the code in subr.c, despite being #ifdef'd out, is still referred to from the Lisp LLSUBR which is in the system images.   I'm fine with removing the commented out CLX options in the makefile fragments, but not the rest of it.

hjellinek:
To be clear, @nbriggs, you're recommending keeping subr.c **and** socket.c, is that right?

Anzus:
I suspect that @BobBane4IL might know more about this than I do, but I don't think XMAS was stable enough to be released?

nbriggs:
I'm suggesting that removing the commented out CLX things in the makefile fragments is OK, but that socdvr.c and socket.c and the commented out code in subr.c should stay.   Even if socdvr.c and socket.c were to go, having the traces of the subrs in subr.c is good up until the point that the SUBR definitions for those calls are removed from LLSUBR.

waywardmonkeys:
What constants are being used by the lisp code?

they have been removed and / or reassigned on the C side. Where the CLX code was using `0244`, this has now been assigned to `sb_MNW_OP`

nbriggs:
```
	(OPEN-SOCKET 160)
        (CLOSE-SOCKET 161)
	(READ-SOCKET 162)
        (WRITE-SOCKET 163)
```
0244 (164) is not in the list of SUBR numbers.

nbriggs:
I think we can see the ripples in history here -- there was an attempt to do native X windows using CLX as the client library, and the subrs provided the interface from the CLX lisp code to the C code that had the connection to the X server.  That was abandoned, for reasons unknown to me, but people were working on it in 1988 or so.  There's code that's currently not included in the emulator build, "mnxmeth.c", (that would be "medley native X methods") which at first glance maps to the methods windows and display streams should implement.  Bob and others were working on native windows code in 1992.  I don't know how XMAS and the MNX C code are related.

Anzus:
OK, I dug through some old emails and I have obviously forgotten more than I care to admit. There was an XMAS2 as late as ’91, done under contract to Fuji Xerox, that “worked” in Medley 2. This had some problems, because it basically implemented Medley’s window “manager” on top of CLX, while X really wants to have it’s own window manager manage some things that are done in Medley. Eventually Fuji decided to go another way. 

Arun Welch
welch@anzus.com

masinter:
XMAS was never released and I'd put it low priority -- doesn't buy anything that we need. A failed experiment.
Worth a footnote in Interlisp history.

waywardmonkeys:
Rebased this forward to fix conflicts. I'd still like to see this land and this code go away.

It sounds like it never really worked and isn't a high priority for the future. If it were, we'd probably start with some of the same code from the modern / maintained version of CLX.

nbriggs:
I think the "modern" version of these files is also about 22 years old.  That said, I guess we should land this and move on.

masinter:
Forwarding pointer to https://github.com/Interlisp/medley/discussions/99 for anyone who wants to talk more.

PR #91, opened 2020-12-17
fix warning: result of the left shift is undefined because the left operand is negative

nbriggs:
There is no functional difference between the old and new code, but we avoid
complaints from code analysis tools such as clang-tidy


waywardmonkeys:
LGTM.

I wish we could move loop variables to the loop (`i` here...)

nbriggs:
If/when we upgrade to C99 or later that sort of thing will be more pleasant to read.
Aside from things like that that just weren't in the language, a lot of the code is written as it is to adapt to what the compiler would optimize (e.g., int often got better optimization than unsigned), and even then there were big wins from re-processing the generated assembler on the SPARC systems.

PR #92, opened 2020-12-17
Missing return type for make_kbd_instance().

waywardmonkeys:
We don't fill in other missing types here as the code relies
upon missing types for the arguments to the various callbacks.


nbriggs (approved):
LGTM

PR #93, opened 2020-12-17
Remove extra arg to open when loading sysout.

waywardmonkeys:
The `open()` call can take a third argument (for when the mode
is `O_CREAT`). However, it is not needed here as we aren't
using `O_CREAT`. (And if we needed an extra arg, it wouldn't
have been `NULL`.)


waywardmonkeys:
(This is another thing that I found via my usage of `clang-tidy` and my experimental `cmake`-based build system.)

nbriggs (approved):
LGTM

PR #95, opened 2020-12-17
Remove src/colorbltfns.c

waywardmonkeys:
This was renamed into `src/rawcolor.c` when the renaming for DOS
happened.


nbriggs (approved):
... and was subsequently updated.  The original code wouldn't have compiled (defualts: ??)

LGTM.

PR #96, opened 2020-12-17
Remove FSERROR define as it is always on.

waywardmonkeys:
The FSERROR define is on all the time now. It was not controlling
all of the functionality, with many places setting Lisp_errnno
outside of the code controlled by the define.


nbriggs (approved):
LGTM.

PR #97, opened 2020-12-17
The error() function may return under exceptional circumstances

nbriggs:
There is no practical difference between the old and new code, but we avoid
compiler warnings for reaching the end of a non-void function.
Attempting to return to Lisp after entering uRaid from one of these
error() calls will result in undefined behavior in Lisp.

### DIFF
--- a/src/car-cdr.c
+++ b/src/car-cdr.c
@@ -79,6 +79,7 @@ LispPTR car(register LispPTR datum)
       return (NIL);
     else
       error("car : ARG not list");
+    return (NIL); /* NOT REACHED */
   }
 } /* end of car */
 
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I have a git stash around with a different fix for this ... What I'd done instead was to flag `error()` with `__attribute__((__noreturn__))`, via setting up a `MAIKO_NORETURN` macro.

Can `error()` actually return?

It may also be useful to have a `MAIKO_UNREACHABLE()` that invokes gcc / clang's `__builtin_unreachable()` as well in places.

waywardmonkeys:
And FWIW, this `MAIKO_NORETURN`  can be supported by Solaris Studio's compiler:

https://docs.oracle.com/cd/E18659_01/html/821-1384/gjzke.html

nbriggs:
It drops you into uRaid, but if you tell that "return to Lisp" it will return.  It's not clear that you'd have a particularly functional system (gc may be confused, for example, or the free list for cons cells on a page may be damaged), but you might be able to do an emergency save of unsaved work.

PR #98, opened 2020-12-18
Remove sccs id lines

nbriggs:
The (old) version and copyright information is in a comment in each file, and this removes an unused variable warning.


waywardmonkeys:
LGTM

PR #99, opened 2020-12-19
Replace calls to putenv with setenv

PuercoPop:
I've verified that that the code builds and the medley image starts, but I'm not sure how to exercise the code paths call fork. Also I went ahead and removed the calls to malloc, instead reusing the same tempstring across calls to setenv.

Closes https://github.com/Interlisp/medley/issues/83

### DIFF
--- a/src/unixfork.c
+++ b/src/unixfork.c
@@ -269,22 +269,18 @@ int fork_Unix() {
        and put their numbers in the environment so parent can find them */
     /* JDS - NB that sprintf doesn't always return a string! */
 
-    char *tempstring;
+    char tempstring[30];
 
-    tempstring = (char *)malloc(30);
-    snprintf(tempstring, 30, "%d", UnixToLisp[0]);
+    snprintf(tempstring, sizeof(tempstring), "%d", UnixToLisp[0]);
     setenv("LDEPIPEINE", tempstring, 1);
 
-    tempstring = (char *)malloc(30);
-    snprintf(tempstring, 30, "%d", LispToUnix[1]);
+    snprintf(tempstring, sizeof(tempstring), "%d", LispToUnix[1]);
     setenv("LDEPIPEOUT", tempstring, 1);
 
-    tempstring = (char *)malloc(30);
-    snprintf(tempstring, 30, "%ld", StartTime);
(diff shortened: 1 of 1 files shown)


nbriggs:
It feels as though these should be char *buf; asprintf(&buf, ....); free(buf); if there's a printf involved so there's no question about having a long enough buffer.

waywardmonkeys:
I would use `snprintf`, but passing `sizeof(tempstring)` ... it is printing an integer, so the maximum length is well known.

nbriggs:
Except for the termtype, but that's checked as being < 59 characters, so OK.

PuercoPop:
I've updated the code to use sizeof(tempstring).

Regarding termtype, I'm assuming it is being read from the environment variable? If so, given that environment variables are 'user input' shouldn't we use strnlen instead?

nbriggs:
OK.   The termtype is read into a stack allocated buffer of 32 characters, where the length that it actually reads into that buffer is provided by the process at the other end of the pipe, and what's coming across the pipe is user provided input in Lisp, and is not length checked.

-- Nick

PuercoPop:
> termtype is read into a stack allocated buffer of 32 characters

So checking that it is less than 59 is a mistake, right? What do you think doing?

```patch
modified   src/unixfork.c
@@ -147,7 +147,7 @@ int ForkUnixShell(int slot, char ltr, char numb, char *termtype, char *shellarg)
        configure the shell appropriately, though this may not be so important any more */
     setenv("LDESHELL", "YES", 1);
 
-    if ((termtype[0] != 0) && (strlen(termtype) < 59)) { /* set the TERM environment var */
+    if ((termtype[0] != 0) && (strnlen(termtype, 32) != 32)) { /* set the TERM environment var */
       setenv("TERM", termtype, 1);
     }
     /* Start up csh */
@@ -541,4 +541,3 @@ int fork_Unix() {
```

waywardmonkeys:
I'd rather see that handled in a separate PR and get this one for `putenv` itself landed.

nbriggs (approved):
LGTM.  We can move forward from here.

waywardmonkeys:
Thanks @PuercoPop !

PR #100, opened 2020-12-19
integer multiply overflow wasn't catching all cases

masinter:
I thought this had to do with 64-bit, but I don't know C well enough to say.
This fixes the bug that (cl:* 300000 300000) returned a negative number.

### DIFF
--- a/src/arith4.c
+++ b/src/arith4.c
@@ -61,7 +61,7 @@ dummy:
 #else
 
   result = arg1 * arg2;
-  if ((arg2 != 0) && ((result / arg2) != arg1)) goto doufn2;
+  if ((arg2 != 0) && (((result % arg2) != 0) || ((result / arg2) != arg1))) goto doufn2;
   N_ARITH_SWITCH(result);
 
 #endif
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM

nbriggs:
The two modes we compile in are ILP32 (ints, longs, pointers are all 32-bit) and LP64 (longs and pointers are 64-bit, ints remain 32-bit).

nbriggs:
Hmmm... it's better, but not complete.  Test it with 65536 * 65536 (or any product of powers of 2)

masinter:
did the old code work anywhere? I confess to not having tested thoroughly. It took me the better part of the day just tracking down the problem, and I worry a little about there being Interlisp code that expects ignore-overflow behavior.  I can back up and try again.

waywardmonkeys:
This patch shouldn't be applied. I'll explain in more detail soon.

waywardmonkeys:
Long explanation with C and x86_64 assembler given in medley/issues#90, along with my proposal for the next 4 steps.

PR #103, opened 2020-12-20
Update how we configure gcc / clang command lines.

waywardmonkeys:
This PR centralizes the configuration of clang and gcc command lines, and then in a subsequent commit, adds `-fwrapv` to deal with the problems discussed in interlisp/medley#90.


nbriggs (approved):
LGTM.  I think we can dump the "-fno-omit-frame-pointer -Wall -Wextra -fno-aggressive-loop-optimizations" etc. that were experiments as well.

PR #104, opened 2020-12-20
Fix timer.c compilation on non-x86_64 GNU/Linux.

waywardmonkeys:
`timer.c` uses `F_SETSIG` which is a GNU extension. As such, it
needs `_GNU_SOURCE` to be defined, but this was only done for
x86_64, not the other Linux platforms that we have build system
support for.

Closes interlisp/medley#97.


nbriggs (approved):
LGTM, apart from preferring that generally we avoid GNU specific extensions when possible.

waywardmonkeys:
I don't really know why we do this `F_SETSIG` here ... but it has been there, and I haven't figured out how to remove it yet.

waywardmonkeys:
And the `iotcl(..., I_SETSIG, ...)` is part of STREAMS, I think.

nbriggs:
Yes, I_SETSIG is defined in https://pubs.opengroup.org/onlinepubs/9699919799/

If KBINT is defined then the I/O is going to be driven by the arrival of a keyboard interrupt or an ethernet packet -- it's not going to call the <code>select(...)</code> [with timeout 0] for I/O except on receipt of that signal.  If you're on SunOS 4.x or you don't have (maiko feature) KBINT then it calls what would have been the signal handler (<code>getsignaldata(...)</code>) in the dispatch loop, which is a little ugly, but it's what we're currently doing.  So, if the <code>[I/F]_SETSIG</code> IS functional then we're BOTH handling the signal to do I/O AND calling the same code in the dispatch loop.  It would be interesting to define KBINT as an experiment and see if it still works well.   It does need something to trigger calls to the signal handler so that any other async I/O happens.

masinter:
There's a problem with whatever we're doing now, that leads to run-away lisp code that doesn't get keyboard interrupts. 
I think @pmcjones and @hjellinek both had the experience of typing DIR and getting directory to all levels (bug 1) but secondly you can't type control-D control-E control-P that all are defined in Interlisp to interrupt and they have no effect, or sometimes not for minutes.

nbriggs:
I don't believe this is a new problem, and I don't believe it has anything to do with the timer.c changes.

The dispatch loop calls getsignaldata(...) that will notice X events including keyboard input (we should be able to trace this), but the code that enumerates directories doesn't cope well with directories with large numbers of files.  It's trying to construct the complete list of files according to Interlisp naming rules, and it has to enumerate them *all* to the depth you specify before it can get back to the dispatch loop.

masinter:
is it time to explore the twisty passages of DSK.c?
I don't know what is the use case for enumeration to any level other than 1. Or what happened to the design rule that lisp threading required no subr or opcode could take more than 1/60th of a second.

PR #105, opened 2020-12-20
Update FreeBSD makefile to use clang and new style for configuring options

nbriggs:
Followup to PR #103. FreeBSD has installed clang by default since 10.0-CURRENT and 9.0-STABLE and gcc is no longer installed by default.  Update the makefile fragment to use clang and the default CLANG_CFLAGS as per other platforms.


waywardmonkeys:
LGTM

PR #106, opened 2020-12-21
Fix typo #ifdef RESWAPPEDDCODESTREAM -- should be RESWAPPEDCODESTREAM

nbriggs:
All uses of the defined symbol, which are intended to be the same symbol, must be spelled in the same way.


waywardmonkeys:
The defined symbol being fixed here is actually `RESWAPPEDCODESTREAM`, so the commit text and PR title should be updated to reflect that.

waywardmonkeys:
LGTM

PR #107, opened 2020-12-21
ForkUnixShell: Replace strlen with strnlen

PuercoPop:
We also reduce the upper bound length to 32 as termtype is defined as
char termtype[32]. strlen may result in reading past the memory
allocated for termtype if no null byte is found in the first 32 bytes,
Use strnlen to avoid that.

This is a follow up to what was mentioned on https://github.com/Interlisp/maiko/pull/99#issuecomment-748418032 Let me know if I missed something and we should be keep 59 bytes as the upper limit.

### DIFF
--- a/src/unixfork.c
+++ b/src/unixfork.c
@@ -229,7 +229,7 @@ int fork_Unix() {
   pid_t pid;
 
   char IOBuf[4];
-  unsigned short tmp;
+  unsigned short tmp = 0;
   char *cmdstring;
 
   /* Pipes between LISP subr and process */
--- a/src/unixfork.c
+++ b/src/unixfork.c
@@ -317,7 +317,7 @@ int fork_Unix() {
       case 'S':
       case 'P':          /* Fork PTY shell */
         if (slot >= 0) { /* Found a free slot */
-          char termtype[32];
+          char termtype[64];
 #ifdef FULLSLAVENAME
           char slavepty[32]; /* For slave pty name */
 
--- a/src/unixfork.c
+++ b/src/unixfork.c
@@ -146,7 +146,7 @@ int ForkUnixShell(int slot, char ltr, char numb, char *termtype, char *shellarg)
        configure the shell appropriately, though this may not be so important any more */
     setenv("LDESHELL", "YES", 1);
 
-    if ((termtype[0] != 0) && (strlen(termtype) < 59)) { /* set the TERM environment var */
+    if (termtype[0] != 0) { /* set the TERM environment var */
       setenv("TERM", termtype, 1);
     }
     /* Start up csh */
@@ -327,7 +327,9 @@ int fork_Unix() {
 
           if (IOBuf[0] == 'P') { /* The new style, which takes term type & command to csh */
             if (SAFEREAD(LispPipeIn, (char *)&tmp, 2) < 0) perror("Slave reading cmd length");
+            if (tmp > 63) perror("Slave termtype length too long");
             if (SAFEREAD(LispPipeIn, termtype, tmp) < 0) perror("Slave reading termtype");
+            termtype[tmp] = '\0';
             if (SAFEREAD(LispPipeIn, (char *)&tmp, 2) < 0) perror("Slave reading cmd length");
             if (tmp > 510)
               cmdstring = (char *)malloc(tmp + 5);


nbriggs:
Let's look at all the code --
```
    switch (IOBuf[0]) {
      case 'S':
      case 'P':          /* Fork PTY shell */
        if (slot >= 0) { /* Found a free slot */
          char termtype[32];
#ifdef FULLSLAVENAME
          char slavepty[32]; /* For slave pty name */

          if (SAFEREAD(LispPipeIn, (char *)&tmp, 2) < 0) perror("Slave reading slave pty len");
          if (SAFEREAD(LispPipeIn, slavepty, tmp) < 0) perror("Slave reading slave pty id");
#endif /* FULLSLAVENAME */

          if (IOBuf[0] == 'P') { /* The new style, which takes term type & command to csh */
            if (SAFEREAD(LispPipeIn, (char *)&tmp, 2) < 0) perror("Slave reading cmd length");
            if (SAFEREAD(LispPipeIn, termtype, tmp) < 0) perror("Slave reading termtype");
            if (SAFEREAD(LispPipeIn, (char *)&tmp, 2) < 0) perror("Slave reading cmd length");
            if (tmp > 510)
              cmdstring = (char *)malloc(tmp + 5);
            else
              cmdstring = shcom;

            if (SAFEREAD(LispPipeIn, cmdstring, tmp) < 0) perror("Slave reading shcom");
          } else /* old style, no args */
          {
            termtype[0] = 0;
            cmdstring = shcom;
            cmdstring[0] = 0;
          }
```
It's allocating 32 bytes on the stack then it reads a 2-byte value for a length (0-65535), and then that many characters into the 32 byte termtype.  That's wrong.  As with the next couple of lines it should check that the length provided is not greater than the memory allocated.  We should consider what a likely maximum size of the termtype is-- looking at an /etc/termcap we find that the maximum name (including aliases) is 61, so perhaps an appropriate max length would be 64 (63+null).  This code should ensure that either there's a properly terminated name, or the first character should be 0.

Now, this code calls ForkUnixShell passing the termtype, that value is going to get passed to the setenv at
```
    if ((termtype[0] != 0) && (strlen(termtype) < 59)) { /* set the TERM environment var */
      setenv("TERM", termtype, 1);
    }
```
So why 59? Remember that it used to construct "TERM=xxxx" to pass to putenv, that looks like 64 bytes less 5 for TERM= and one  extra for the terminating null.  I'm betting at some point termtype was [64] not [32].

So, what should we do?   I think that as far as termtype goes (ignoring slavepty for now, it should probably get the same treatment -- I haven't looked at what could be sent in that value) we should allocate termtype[64] on the stack, read the length, if it's < 64 then use it, ensuring that the name is null-terminated (or zero the first byte), then pass it to ForkUnixShell, and just use it -- there's no reason to recheck the length before passing it to setenv

PuercoPop:
@nbriggs Thanks for the detailed analysis. I've updated the code to:

1. Increase the size of termtype to 64.
2. Check the length with which `read` would be called when using termtype as a buffer and abort if the length is greater than 63, to account for the null byte.
3. Set the null byte assuming that amount of bytes read was equal to the count argument (stored in tmp) with which read was called with. 

This will avoid `setenv` potentially going past the allocated memory when setting termtype. However it is possible that `read` reads less bytes than the amount we ask it to. In that case setenv could use uninitialized memory when setting the value. Let me know if you'd prefer we use the return value of SAFEREAD to set the null byte more precisely.

Another thing that I've noticed is that we are guarding against`SAFEREAD` returning a negative number at every call-site. This seems redundant as inside `SAFEREAD` we are checking if read returns < 0 and either we call read again or error out (through `perror`) so it doesn't seem possible for `SAFEREAD` to return anything other than the amount of bytes read by `read`. Am I missing something?

nbriggs:
```
SAFEREAD(int f, char *b, int c)
{
  ssize_t res;
loop:
  res = read(f, b, c);
  if ((res < 0)) {
    if (errno == EINTR || errno == EAGAIN) goto loop;
    perror("reading UnixPipeIn");
  }
  return (res);
}
```
<code>perror()</code> returns after printing the message, so <code>SAFEREAD</code> can return a negative result, and the additional check at the call site is necessary, and it should probably recover properly from a failure instead of just pressing on.

<code>tmp</code> should be initialized before it is read into, because an error reading into <code>tmp</code> could leave 1 or 2 bytes uninitialized, or <code>tmp</code> could have the last set value.

PuercoPop:
> tmp should be initialized before it is read into, because an error reading into tmp could leave 1 or 2 bytes uninitialized, or tmp could have the last set value.

Done in https://github.com/Interlisp/maiko/pull/107/commits/6eea8601000dc578bdc009c15928282254573204

nbriggs (approved):
LGTM.

nbriggs:
Merged by hand to resolve conflicts.   Thanks!

PR #109, opened 2020-12-21
Remove dead cursor code

nbriggs:
Code in xcursor.c initialized various cursor image related variables that are no longer used.


waywardmonkeys:
LGTM (visual check only).

PR #110, opened 2020-12-21
Remove display name handling block ifdef'd for X11R1.

waywardmonkeys:
X11R1 was released in 1987 and superseded by X11R2 in 1988.

We should be able to safely remove this code. We set the
preprocessor variable on all platforms to indicate that we're
on X11R4 (or later, I guess).


nbriggs (approved):
LGTM.  Oddly, the code controlled by the #ifdef seems unrelated to the X version, rather it's about whether the option was named or not.

PR #113, opened 2020-12-22
Remove XV11R4 and related defines.

waywardmonkeys:
We no longer have anything predicated upon the version of X11 that
we're using (and these versions are all ancient anyway).


nbriggs (approved):
LGTM.

PR #115, opened 2020-12-22
Remove empty, unreferenced bbt386i.h.

waywardmonkeys:


nbriggs:
I'm left wondering how in the Unix to DOS renaming --
<code>./unix2dos.sed:s/bitblt386i\.s/bbt386i\.h/g</code> we ended up with the assembler source becoming a .h file, and then getting lost.  Not good.

nbriggs (approved):

PR #118, opened 2020-12-22
Include termios as <termios.h>, not <sys/...>.

waywardmonkeys:
termios is available via `<termios.h>` on all supported platforms,
so variants of `<sys/termio.h>` and `<sys/termios.h>` are no
longer needed.


nbriggs (approved):
LGTM.

PR #119, opened 2020-12-22
Remove call to ieee_handler().

waywardmonkeys:
When `FLTINT` is defined (which is currently only on SunOS), if
`OS5` was not defined, then it tried to use `ieee_handler` which
doesn't exist on Linux and friends. If we were to enable `FLTINT`
on more platforms, then we'd want to always be using more modern
signal handling and not `ieee_handler`.


nbriggs (approved):
LGTM.

PR #120, opened 2020-12-22
Remove ToMem macro.

waywardmonkeys:
We can always use `memcpy` to copy from a `hostent` struct to a
`sockaddr_in` struct as they will not be overlapping.


nbriggs (approved):
LGTM.

PR #122, opened 2020-12-22
Remove prototypes for getenv. Use stdlib.h.

waywardmonkeys:
This also removes some casts to `char *` that were present but
not needed, since `getenv()` is defined to return `char *`.

### DIFF
--- a/src/ldeether.c
+++ b/src/ldeether.c
@@ -63,6 +63,7 @@ char *devices[] = {"le0",   "le1",   "le2",   "le3",   "le4",   "ie0", "ie1", "i
 #include <nlist.h>
 #include <fcntl.h>
 #include <malloc.h>
+#include <stdlib.h>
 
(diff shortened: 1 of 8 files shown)


src/ldeether.c
```diff
@@ -103,7 +103,7 @@ int main(int argc, char *argv[]) {
     /* Use DLPI to connect to the ethernet.  This code is stolen
        from NFSWATCH4.3
     */
-    char *etherdev = (char *)getenv("LDEETHERDEV");
+    char *etherdev = getenv("LDEETHERDEV");
```
nbriggs, line 106: Missing the <code>#include <stdlib.h></code>

waywardmonkeys: Fixed and force pushed.

nbriggs (approved):
LGTM

PR #124, opened 2020-12-22
Remove XWINDOWS define from FreeBSD, Linux, and macOS builds.

waywardmonkeys:
The right define is `XWINDOW` which is already defined via the
build system.


nbriggs (approved):
LGTM

PR #126, opened 2020-12-22
Remove comment reference to SIGNED_CHARS define.

waywardmonkeys:
None exists or is used, so no need to discuss it.


nbriggs (approved):
Signed characters used to be an issue, and the code should probably be clearer where it is depending on the signedness of <code>char</code>.  A lot of the usage is really as <code>int8_t</code> and <code>uint8_t</code> types and if we get to the point of being C99 compliant (or later) there are a lot of changes to make in that regard.

PR #127, opened 2020-12-22
Clean up medleyfp.h.

waywardmonkeys:
* Remove sections for sparc and Sun i386 for non-Solaris.
* Have Linux use the same code as FreeBSD and macOS.
* Make the FreeBSD / Linux / macOS code path the default, if nothing
  special is defined since it is the modern way.
* Remove fpstatus_() usage from the ancient days.

### DIFF
--- a/inc/medleyfp.h
+++ b/inc/medleyfp.h
@@ -27,7 +27,6 @@
 #ifdef FLTINT
 volatile extern int  FP_error;
 
-
 /*  Note that a compiler may very likely move code around the arithmetic
     operation, causing this test (set by an interrupt handler) to be
     incorrect.  For example, the Sun SPARC compiler with -O2 makes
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I haven't tested this with a running Medley image. It might be good to do that.

nbriggs:
On Solaris, <code>finite()</code> is defined in ieeefp.h.
```
cc -m32 -funsigned-char -O2 -c -O2 -g3 -DXWINDOW  -DOS5 -DUSE_DLPI -DNOPIXRECT -DSYSVSIGNALS -DOLD_CURSOR -DNEW_STORAGE -DLOGINT -DFORKCOMM -DLOCK_X_UPDATES  -DRELEASE=351  -I../inc/ -I../include/ ../src/fp.c  -o ../sunos5.sparc-x/fp.o
"../src/fp.c", line 63: warning: implicit function declaration: finite
```

waywardmonkeys:
Odd. That wasn’t included before and I don’t think I changed the Solaris part...

nbriggs:
It has probably been wrong for a long time.  Still, no reason not to fix it now.

There's a very subtle difference, but it may not actually be detectable in the result here, between <code>finite(double dsrc)</code>, which should cause promotion from float to double for our usage, and <code>isfinite()</code> which is a macro and is specifically defined as not promoting the argument type.

waywardmonkeys:
Ah, I see. I will fix it after I sleep. I wanted to use the same code path on Solaris as other Unix platforms, but you had indicated that that didn’t work.

waywardmonkeys:
Pushed an update to add `<ieeefp.h>` on Solaris.

nbriggs (approved):
LGTM. Checked compile on Solaris.

PR #128, opened 2020-12-22
Don't use dirent's d_reclen as if it were d_namlen.

waywardmonkeys:
This code was treating the `d_reclen` incorrectly from some of the
original old porting work. It is not the same as `d_namlen` and
`d_namlen` is not widely supported or standard. It should be using
`strlen(d_name)`.

This addresses an issue mentioned in interlisp/medley#103.


nbriggs (approved):
LGTM.

PR #129, opened 2020-12-23
Use uname() for getparm for ARCH/MACH queries.

waywardmonkeys:
Previously, this was implemented in very inconsistent ways and
not yet implemented for most of the platforms that we're supporting
today.

The values for DOS are maintained as they were.


nbriggs:
It's not clear to me why returning the OS name (sysname) for ARCH is a good choice.  The existing code was, I suspect, attempting to duplicate the result of the SunOS/Solaris "arch" and "mach" commands and simulate it where it didn't exist.  I wouldn't object to OSNAME as another parameter it returns, though, because it makes it easler for Lisp code to adapt to the underlying system.

|system|arch|mach|uname -p|uname -i|uname -m|
|-----|-----|-----|-----------|-----------|-----------|
|Solaris 11.4 on a T1|sun4|sparc (sysinfo(SI_ARCHITECTURE))|sparc|sun4v|sun4v|
|Solaris 11.4 (64-bit) on VirtualBox on Intel(R) Core(TM) i7-8850H|i86pc|i386|i386|i86pc|i86pc|
|Linux on Intel(R) Core(TM) i7-8850H|x86_64|(none)|x86_64|x86_64|x86_64|
|Linux on ARMv7 TI AM335x BeagleBone Black|armv7l|(none)|unknown|unknown|armv7l|
|macOS 10.11 on Intel core 2 duo|i386|(none)|i386|(none)|x86_64|
|mac OSX on Power PC|ppc|(none)|powerpc|(none)|Power Macintosh|
|FreeBSD 12.2 on IntelPentium M processor (686-class CPU)|(none)|(none)|i386|GENERIC|i386|

FreeBSD says

> make(1) uses [uname -p] to set the MACHINE_ARCH variable.
> make(1) uses [uname -m] to set the MACHINE variable.

Solaris 11.4 says:

> On sparc systems,` arch` returns sun4 (historical artifact)
> while `arch -k` returns `uname -m`. On all other systems,
> `arch` == `arch -k` == `uname -m`.

masinter:
what is most useful going forward? Labels to put on benchmarks? deciding what to send to ShellCommand?

waywardmonkeys:
How is this actually used from within Medley? The comment indicates that it isn’t used much except for whether or not `ARCH` is `”dos”` or not and whether or not there is any result from the call ... is that still accurate?

nbriggs:
It's not a closed universe -- we have no idea how it might have been used.

waywardmonkeys:
The good thing about a 30 year gap in history is that all of the platforms that were supported, except Solaris and sort-of-DOS, are dead now. So any legacy code checking this for things like MIPS or 68k or Irix or AIX ... wouldn't care what the new values are. So I think it is best if we can get it right now. It is also worth noting that many of our current platforms are mis-reported by this, like any ARM / Aarch64 platforms.

masinter:
getting something that is meaningful for labelling benchmarks would be good.

masinter:
doesn't this provide Ron with what he needs for clipboard?

nbriggs:
No, it doesn't tell you which OS you're running -- could be x86_64 darwin, or x86_64 linux, and they need different copy/paste commands.   It also tells you "X", but they're all "X', so that's no help.   I think we can probably settle on something for "MACH" and "ARCH", but Ron needs "OSNAME" -- and we'll get that from the OS we compiled on.  I've got a lot of things I'm working on.

waywardmonkeys:
Ron should be using new subrs that let Maiko correctly implement copy/paste. Doing otherwise simply won't work on all platforms going forward as shell commands aren't always available to do this.

PR #130, opened 2020-12-23
Simplify including sun/fbio.h.

waywardmonkeys:
Both sides of this #ifdef were the same, so just collapse it.

This code isn't actually relevant any longer, but this removes
an OS-specific #ifdef, making inspection of platform specific
code a bit easier.


nbriggs (approved):
LGTM

PR #131, opened 2020-12-23
Remove bzero macro def and usages.

waywardmonkeys:
We can just use memset for bzero and we weren't using bcmp at all.

### DIFF
--- a/inc/version.h
+++ b/inc/version.h
@@ -289,8 +289,6 @@ typedef unsigned char u_char;
 typedef unsigned long u_int;
 typedef signed char s_char;
 typedef unsigned short u_short;
-	/* DOS doesn't have the BSD bzero &c functions */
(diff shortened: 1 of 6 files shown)


nbriggs:
inc/os.h and src/dlpi.c go together and are almost verbatim copies from nfswatch (https://sourceforge.net/projects/nfswatch/files/nfswatch/4.99.11/ last released 2010).  While I think it's great to stop using bzero/bcopy (as you note, we don't use bcmp) in our own code, I'm not sure that it's worthwhile cleaning up os.h & dlpi.c, and we don't include os.h in any of our own code.

waywardmonkeys:
Okay, updated the patch to no longer modify `inc/os.h`.

nbriggs (approved):
LGTM

PR #134, opened 2020-12-26
Remove UNSAFE flag.

waywardmonkeys:
This flag was for SunOS 3 and 4 on sun3 hardware (68k).

It enabled some unsafe optimizations via assembler. This hasn't
actually been built since the DOS port (since the header for this
was renamed, but the code in `xc.c` wasn't updated).

The optimizations here made assumptions that were specific to the
68k hardware and register allocation. They aren't likely to be
of much use moving forward, and we've got plenty of safe performance
improvements from the last 30 years of hardware advances.


waywardmonkeys:
This and #133 are things that I'm looking at as I work towards some other changes for the use of `__builtin_*_overflow()` for math ... I want to clean up and make these code paths a lot easier to follow, once we don't need custom assembler for the optimizations.

nbriggs (approved):
LGTM

PR #135, opened 2020-12-27
NEW_STORAGE is no longer new.

waywardmonkeys:


nbriggs (commented):
I *think* this is OK -- the only hesitation is because the original makefile-init.* files (which don't include the makefile-init.386) all did NOT have -DNEW_STORAGE. Looking in the Lisp side of things for the Interface Page we find in LLPARAMS:
```
             (EMUBUFLENGTH WORD)  (* ; "Number of words of said space")
       (* ;;  "The following 2 are available if NEW_STOARGE is specified in C")
             (ProcessSize WORD)  (* ; "Process size for which can be use as LISP space")
             (StorageFullState WORD) (* ; "Save last storage state")
             (ISFMAP WORD)
```
So we need to leave tracks so people reading the Lisp side know that those two words are now always there.  Oddly, when I looked in the LLBFS (low level basic file system) code, where it would use the "sysdisk" field that is being replaced by "storagefullstate", I found that I was the last person who edited <code>\INITBFS</code>, 16 Feb 1987 10:26 AM.

waywardmonkeys:
How should we do this on the Lisp side?

nbriggs:
We can't remake the LLPARAMS file at the moment, so perhaps we should leave a comment in the version.h saying that <code>-DNEW_STORAGE</code> was removed as it is now always true and that that is relevant to the comments in LLPARAMS

waywardmonkeys:
I've added a comment in `inc/version.h`, rebased against current `master`, and force-pushed.

Does the comment in `inc/version.h` look okay to you?

nbriggs (approved):
LGTM.

PR #136, opened 2020-12-28
VM config controlled by RELEASE version

waywardmonkeys:
This is a placeholder PR that requires #135 to be merged first. Then I'll update this one as needed.


nbriggs (commented):
BIGVM and NEWCDRCODING are candidates for similar cleanup.  They are both defined in version.h with respect to particular RELEASE values (though your commit comment says VERSION)

waywardmonkeys:
I was going to do those separately, but I will combine them into a single PR (this one).

waywardmonkeys:
Updated and pushed. This just shows though that a bunch of the DOS-related files in `bin` are ancient relics and need to be removed. I wish we knew why half those files even existed.

nbriggs:
Some should certainly go, next pass, although they're not currently doing any harm.
makefile-dos and makefile.dos differ in whether they refer to /src or /dossrc and /inc vs /dosinc, and one of them had the Linear Programming solver (an external package) built in (-DLPSOLVE)

mkfile.i looks like what you get if you run mkfile.c through the C preprocessor with a few -Dxxx values, and mkfile is a further processed version, i'd guess.  It looks as though Jarl Nilsson was reworking the build system.  I guess he used cpp rather than m4 because it got run on DOS boxes :-(   Glad I didn't have anything to do with it.

nbriggs (approved):
LGTM

PR #137, opened 2020-12-28
init_Xcursor takes Window, not int for window.

waywardmonkeys:
`Window` is an X11 typedef for `unsigned long` (via `XID`). We should
pass it as that, not an `int`, to avoid 64 bit conversion issues.


nbriggs (approved):
LGTM

PR #139, opened 2020-12-28
Use `SEEK_SET` for third arg to `*seek()` functions.

waywardmonkeys:
Instead of sometimes passing `0` to `lseek` and friends, we should
use the standard symbolic constant `SEEK_SET`.


nbriggs (approved):
LGTM

PR #141, opened 2020-12-28
fcntl(..., F_SETFL, ...) return value only signals error.

waywardmonkeys:
The return value of `fcntl` when passed `F_SETFL` isn't guaranteed
to return the flags passed in as its return value. It will be `-1`
in the event of an error, but no other value is to be relied upon.

Fortunately, we weren't relying upon the value apart from
occasionally checking for an error.

This is as it is specified in https://pubs.opengroup.org/onlinepubs/7908799/xsh/fcntl.html

Closes interlisp/medley#87.

### DIFF
--- a/src/Cldeetr.c
+++ b/src/Cldeetr.c
@@ -142,7 +142,7 @@ int main(int argc, char *argv[])
       strcpy(Ename, if_data.ifc_req[0].ifr_name);
 
       flags = fcntl(ether_fd, F_GETFL, 0);
-      flags = fcntl(ether_fd, F_SETFL, flags | FASYNC | FNDELAY);
+      fcntl(ether_fd, F_SETFL, flags | FASYNC | FNDELAY);
 
 #ifdef DEBUG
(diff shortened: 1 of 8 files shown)


nbriggs (commented):
This might be a good time to move consistently to O_NDELAY (from FNDELAY) and O_ASYNC (from FASYNC) since those are the POSIX names and the others are just included for compatibility.  Also, we could switch style to
```
     [res =] fcntl(fd, F_SETFL, fcntl(fd, F_GETFL, 0) | O_NDELAY);
...
     [res =] fcntl(fd, F_SETFL, fcntl(fd, F_GETFL, 0) & ~O_NDELAY);
```
(I'm not a fan of 65535 - FNDELAY)

waywardmonkeys:
I have a separate bug for that change and plan to do it as a separate PR. I want to keep the changes here limited to fixing the one thing.

nbriggs (approved):
LGTM.

PR #142, opened 2020-12-29
Use gcc / clang overflow builtins.

waywardmonkeys:
This avoids expensive checks for overflow that employ undefined
behavior.

This is a step along the way towards replacing the old hand-written
assembler that did the same thing in terms of using the CPU's
overflow detection.


waywardmonkeys:
This is a work in progress branch of mine. It can't use `__has_builtin` reliably as it isn't available until more recent versions of gcc.

waywardmonkeys:
@nbriggs This is getting closer to where I'm happy with it.

nbriggs (approved):
LGTM.  Tested on SPARC with gcc 10.2.0, verified that it generated the __builtin_... function calls.

PR #144, opened 2020-12-29
Update fcntl F_SETFL usage.

waywardmonkeys:
This changes from `FASYNC` to `O_ASYNC`, `FNDELAY` to `O_NONBLOCK`,
and `O_NDELAY` to `O_NONBLOCK`. These are the modern names.

`O_NONBLOCK` is part of the POSIX standard. However, `O_ASYNC` is
specific to Linux and BSD. It is not available on Solaris, where
the `I_SETSIG` `ioctl` should be used instead, or things should
be done some other way, with `poll()` or the like. Also, the
behavior of having I/O trigger a `SIGIO` signal is not in POSIX,
since the `SIGIO` signal is not in POSIX. Instead, it is only
the behavior of having `SIGURG` being signalled for out of band
data that is specified.

We also takes this opportunity to collapse some multi-line calls
to get the flags, store it into a temp, and then set them, to
just doing it in one line, skipping the stored temporary value.

We also change one instance of `65535 - FNDELAY` to `~O_NONBLOCK`.

Closes interlisp/medley#85.

### DIFF
--- a/src/Cldeetr.c
+++ b/src/Cldeetr.c
@@ -50,7 +50,6 @@ int main(int argc, char *argv[])
 {
   char Earg[30], Ename[30], **newargv;
   int i;
-  int flags;
   /* Kickstart program for the Lisp Development Environment (LDE).
           Run this as setuid root to open the LDE ether socket.
           Passes all arguments through to LDE plus -E <ether-info>
@@ -141,8 +140,7 @@ int main(int argc, char *argv[])
       bcopy(if_data.ifc_req[0].ifr_addr.sa_data, ether_host, 6);
       strcpy(Ename, if_data.ifc_req[0].ifr_name);
 
-      flags = fcntl(ether_fd, F_GETFL, 0);
-      fcntl(ether_fd, F_SETFL, flags | FASYNC | FNDELAY);
+      fcntl(ether_fd, F_SETFL, fcntl(ether_fd, F_GETFL, 0) | O_ASYNC | O_NONBLOCK);
 
 #ifdef DEBUG
       printf("init_ether: **** Ethernet starts ****\n");
--- a/src/chardev.c
+++ b/src/chardev.c
@@ -84,7 +84,7 @@ LispPTR CHAR_openfile(LispPTR *args)
   Lisp_errno = (int *)(Addr68k_from_LADDR(args[2]));
 
   LispStringToCString(args[0], pathname, MAXPATHLEN);
-  flags = O_NDELAY;
+  flags = O_NONBLOCK;
   ERRSETJMP(NIL);
   /*    TIMEOUT( rval=stat(pathname, &statbuf) );
       if(rval == 0){      } */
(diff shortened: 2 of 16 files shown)


waywardmonkeys:
@nbriggs Could you see if this even compiles on Solaris since apparently there isn't `OASYNC` there? I'm not sure why it compiled with `FASYNC` either, but ...

Also, there's at least one other likely bug here where 2 of the socket cases in `src/inet.c` are setting `O_NONBLOCK`  but not `O_ASYNC`, despite doing a `F_SETOWN` on the socket. This is likely an omission where it should be setting both `O_NONBLOCK` and `O_ASYNC`. That should be handled in a separate PR though.

nbriggs:
Yeah, problem on Solaris --
```
"../src/inet.c", line 109: undefined symbol: O_ASYNC
$ ggrep -r -e O_ASYNC -e FASYNC /usr/include
/usr/include/sys/file.h:#define FASYNC          0x1000  /* asyncio in progress pseudo flag */
```
Also odd, but I suspect there is no consequence for using <code>O_NONBLOCK</code> rather than <code>O_NDELAY</code> where it used to be <code>FNDELAY</code>.
```
ggrep -r -e O_NONBLOCK -e O_NDELAY /usr/include/sys/fcntl.h
#define O_NDELAY        0x04    /* non-blocking I/O */
#define O_NONBLOCK      0x80    /* non-blocking I/O (POSIX) */
```
I haven't looked into the <code>F_SETOWN</code> issue there.

nbriggs:
Link to docs regarding [interrupt driven I/O on Solaris 11.2](https://docs.oracle.com/cd/E36784_01/html/E36861/sockets-82255.html#scrolltoc) with reference to the Asynchronous Socket I/O as well, circa 2014.

waywardmonkeys:
1) Is `inet.c` the only file that fails to compile on Solaris?

2) There is a difference between `O_NDELAY` and `O_NONBLOCK`, but primarily only on Solaris. This goes back into ancient history, but it is covered here: https://stackoverflow.com/questions/1150635/unix-nonblocking-i-o-o-nonblock-vs-fionbio

3) The missing `O_ASYNC` stuff will be done in a separate PR ... I'll probably file a bug so that I don't forget when I inevitably get distracted.

nbriggs:
Yes, <code>inet.c</code> is the only file that failed to compile on Solaris.

waywardmonkeys:
It looks like I might get away with:

```
#if defined(OS5) && !defined(O_ASYNC)
#define O_ASYNC FASYNC
#endif
```

either in `inc/version.h` or in `src/inet.c`. If we were to ever try to compile on SunOS4, we'd need something similar as well, since when I modified `src/ether.c` and some others, I modified `OS4` blocks.

nbriggs:
Yes, the <code>#define O_ASYNC FASYNC</code> is sufficient to have a compilable and generally working system on Solaris 11 -- However, I don't have a test case to exercise the inet.c code at the moment.  I'd still say go with that, and if we can test the inet code and it fails we can fix it.

waywardmonkeys:
I've updated the commit message and added a define for Solaris.

nbriggs (approved):
LGTM.  Checked for compile and basic operation on Solaris 11.4/SPARC

PR #146, opened 2020-12-30
README: Add discussion of platform support.

waywardmonkeys:


masinter (changes requested):
I want to avoid wording that implies any kind of "support" for anything. We've "tested on", we "believe it will work on", but we haven't tested extensively, and I have reproducible cases where the raspberry pi fails and other platforms don't.

waywardmonkeys:
Updated to change the wording.

masinter (approved):

nbriggs (approved):
LGTM

PR #147, opened 2020-12-30
Fix makefile fragment for linux on arm

nbriggs:
Clean up compiler flags and switch from clang to gcc as the default compiler.
Neither clang nor gcc on arm appear to recognize the -m32 flag.

### DIFF
--- a/bin/makefile-linux.armv7l-x
+++ b/bin/makefile-linux.armv7l-x
@@ -1,4 +1,4 @@
-# Options for Linux, Intel 386/486 and X-Window
+# Options for Linux, ARMv7 and X-Window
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
LGTM. Any particular reason for changing from clang to gcc?

nbriggs:
The BBB I was running on using the normal 4GB uSD card image with Debian 9 comes with gcc but was too full to download clang as well as the maiko git repo.  While I prefer clang over gcc, I think the default for this particular configuration could be gcc if we can make sure it works.

I've now got a 32GB uSD card for the system, and I've expanded the filesystem image into that extra space, so I would have the capability of testing both compilers -- but I haven't done so yet.

PR #148, opened 2020-12-31
REGISTER is for floats.

waywardmonkeys:
According to `inc/version.h`, `REGISTER` is for floats which may
or may not be able to be in registers due to the `FPTEST` macro
implementation. We shouldn't be using it here for non-float
data.


nbriggs (approved):
LGTM.

I'm thinking that most of the "register" storage class specifications on variables are no longer relevant, since today's compilers seem to ignore them for the most part, except for disallowing taking the address either explicitly or implicitly of such a variable.

waywardmonkeys:
Yes. But I didn't want to do a global change to remove it everywhere yet ... although I do get tempted.

PR #150, opened 2020-12-31
`killpg` is POSIX, so we can rely on it now.

waywardmonkeys:
This removes some usages of the SYSVONLY flag and lets us assume
the existence of `killpg`. It also updates a bit of code to accurately
populate the `pgrp` variable using `tcgetpgrp` rather than an ioctl.


waywardmonkeys:
This leaves a single use of `SYSVONLY`, so we are close to being able to remove it!

Also, I think using `killpg` to send `SIGWINCH` in `src/unixcomm.c` can actually go away in a subsequent patch. Setting the terminal size via `TIOCSWINSZ` should send `SIGWINCH` on its own.

nbriggs (approved):
LGTM.

PR #151, opened 2020-12-31
Cygwin doesn't have O_ASYNC, so fix compilation.

waywardmonkeys:
It looks like Cygwin doesn't actually support signal-based I/O,
so while this compiles, it won't be correct and will need
addressing in another way.

### DIFF
--- a/src/inet.c
+++ b/src/inet.c
@@ -28,8 +28,8 @@
 #include <unistd.h>
 #endif /* DOS */
 
-#ifdef OS5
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
@masinter This should fix Cygwin compilation for you.

nbriggs (approved):
LGTM

nbriggs:
@masinter The master branch should be ready for you to recompile on Cygwin now.

PR #152, opened 2020-12-31
Use setsid on all Unix platforms.

waywardmonkeys:
Previously, we used `TIOCNOTTY` on some platforms (FreeBSD, macOS)
and `setsid` on Linux and Solaris. The correct modern way to do
this is to use `setsid` and the BSDs support this, so let's just
do that.


nbriggs (approved):
LGTM

PR #154, opened 2020-12-31
Cleanup byteswap.c a bit

waywardmonkeys:


nbriggs (commented):
Makes me wish that SunOS had taken up BSD's `byteorder(9)` functions, though they weren't in the FreeBSD release until 2003.  

`word_swap_page()` is actually doing `1234` => `2143` => `4321`, touching every word twice.

nbriggs (approved):
LGTM

PR #155, opened 2020-12-31
timer.c: Clean up / simplify includes.

waywardmonkeys:


src/timer.c
```diff
@@ -36,34 +41,16 @@ void (*prev_int_1c)(); /* keeps address of previous 1c handlr*/
...
-
-#ifndef DOS
+#else /* DOS */
 #include <sys/resource.h>
+#include <sys/time.h>
 #endif /* DOS */
 
 #ifdef OS5
 #include <stropts.h>
 extern int ether_fd;
```
nbriggs, line 51: I believe the `ether_fd` should be `#ifndef NOETHER`.  While Solaris is the only OS we currently have without NOETHER I've been thinking on and off about using `/dev/bpf` to replace `/dev/nit` and resurrect the direct network access (for non-IP packets)

nbriggs (approved):

src/timer.c
masinter: Yesterday I had a zoom call with @jdersch to give him a tour of Medley project (following the slides).
They have an IFS implementation which implements PUP either as raw Ethernet packets or encapsulated in UDP.

He's on the interlisp@googlegroups.com list and has been following.

nbriggs: @masinter -- great, it would be motivating to have something useful to talk to.

waywardmonkeys: I've made this `#ifdef USE_DLPI`. That's what the code using `ether_fd` is ... that's a Solaris-specific thing though and is using the STREAMS API. But I'm happy with having them match, so this PR has been amended.

nbriggs: Hmmm, OK in this case -- in general, `ether_fd` can be used `#ifndef NOETHER` and regardless of `USE_DLPI` -- if you have ethernet enabled but don't have STREAMS then it attempts to set up `ether_fd = open("/dev/nit", O_RDWR | O_ASYNC)` (ether.c:859, but there it's OS4 specific).

PR #156, opened 2021-01-01
Cleanup typeof.c

waywardmonkeys:
It isn't clear to me whether or not `N_OP_dtest` is even used. I didn't see anything, but I didn't know if there's some magic somewhere that hooked it into something that would use it from Lisp-side.

### DIFF
--- a/src/typeof.c
+++ b/src/typeof.c
@@ -30,29 +30,6 @@
 
 /************************************************************************/
 /*									*/
-/*			N _ O P _ d t e s t				*/
-/*									*/
-/*	Check for type conformity, else error.				*/
-/*									*/
-/************************************************************************/
-
-LispPTR N_OP_dtest(register LispPTR tos, register int atom_index) {
-  register struct dtd *dtd68k;
-
-  for (dtd68k = (struct dtd *)GetDTD(GetTypeNumber(tos));
-#ifdef BIGVM
-       atom_index != dtd68k->dtd_name;
-#else
-       atom_index != dtd68k->dtd_namelo + (dtd68k->dtd_namehi << 16);
-#endif /* BIGVM */
-       dtd68k = (struct dtd *)GetDTD(dtd68k->dtd_supertype)) {
-    if (dtd68k->dtd_supertype == 0) ERROR_EXIT(tos);
-  }
-  return (tos);
-} /* OP_DTEST END */
-
-/************************************************************************/
-/*									*/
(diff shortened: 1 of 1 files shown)


nbriggs (commented):
I don't believe N_OP_dtest is used.  I think it could be removed.

waywardmonkeys:
I will modify and update this PR in a few hours then! Thanks.

waywardmonkeys:
So, looking at this more, it ends up that opcode `006` is `DTEST`. Now, why doesn't that call `N_OP_dtest`? Because instead, it invokes a macro `DTEST`, which is defined in `inc/inlineC.h`. There's also an `N_OP_DTEST` macro defined in `inc/native.h`, but that's part of `NATIVETRANS` which is all dead code.

For `N_OP_instancep`, the story is, of course, different.

That is opcode `0303`, and the code in `src/xc.c` invokes the `INSTANCEP` macro. This is defined in `inc/inlineC.h` as well, but the definition there basically just invokes `N_OP_instancep`, so the code here in this file is actually still used.

There's another `INSTANCEP` in `src/loopsops.c`, but that's doing something different, I think, and is unrelated.

waywardmonkeys:
I'm inclined to leave this patch as it is .. and then add a new one that makes the code in `inc/inlineC.h` invoke `N_OP_dtest` for the `DEST` macro. Even better, perhaps, would be to remove the macros for these 2 opcodes from `inc/inlineC.h` and just expand small amount of code within `src/xc.c`

There's a lot of possible cleaning up to do in that direction though!

waywardmonkeys:
Okay, updated with changes that remove duplicated code but don't expand things inside `src/xc.c`.

nbriggs:
I think we should have a larger direction setting discussion before changing much in this area -- I believe the code evolved from being all C (not sure if it was in separate files, I suspect so) to having macros for most opcodes that could be expanded either with C code (inlineC.h; some actually inline some calling the older function definitions) or with inline assembler (from one of the inln{386i,68k,PS2,SPARC,dos}.h).

nbriggs:
I guess I should add -- regarding `DTEST` -- that's going to be a hotspot, so they would have avoided the function call overhead and expanded it with a copy of the code in inlineC.h rather than call the original function.

waywardmonkeys:
Okay, I've updated this again then so that it doesn't change `DTEST`, but does clean up `src/typeof.c`.

nbriggs (approved):
OK

PR #160, opened 2021-01-02
Cleanup uraid

waywardmonkeys:


blakemcbride (approved):
Good cleanups!

nbriggs (commented):
GetDTD and GetTypeNumber are macros.  These functions are provided so that when you're in the debugger (as you likely will need to be, if you end up in uraid) you can use them to print  information about the things you're debugging.

I'd prefer not to remove these.

waywardmonkeys:
@nbriggs Updated to remove that commit that removed those functions.

waywardmonkeys:
@nbriggs I addressed your comments. Could this please go in to get rid of these warnings?

nbriggs:
Yeah, sorry, busy today.

PR #161, opened 2021-01-02
dbgtool: Fix warning from incorrect LispPTR* usage.

waywardmonkeys:
`fnobj_lisp` was being treated as a `LispPTR*` when it was just
a `LispPTR`. This requires fixing up the printfs as well since it
is no longer a void* to treat it as `%p`.


waywardmonkeys:
In some places, we print `LispPTR` values as `0x08x` and others as just `%x`. Since this was printing as `%8p`, I went with `0x08x`.

nbriggs:
I'm good with the choice of `0x%08x` in this case.

nbriggs (approved):

PR #163, opened 2021-01-02
Switch to using POSIX pseudoterminals.

waywardmonkeys:
Previously, we were using SysV pseudo-terminals on Solaris and BSD
pseudo-terminals on other Unix platforms. BSD pseudo-terminals have
been deprecated on Linux and are no longer available in some kernel
configurations.

The POSIX API is basically the same as the SysV API, apart from using
`posix_openpt` instead of `open` with `/dev/ptmx`.

Closes interlisp/medley#121.

### DIFF
--- a/inc/unixfork.h
+++ b/inc/unixfork.h
@@ -1,7 +1,2 @@
 int fork_Unix(void);
-#ifdef FULLSLAVENAME
 int ForkUnixShell(int slot, char *PtySlave, char *termtype, char *shellarg);
-#else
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
@masinter Please give this a try and see if it works for you. (I've only built it, not tried running it with a sysout.)

nbriggs (approved):
LGTM.  Checked on macOS 10.11 and Solaris 11.4.

nbriggs:
I'll wait for @masinter's approval before merging, or Larry can do it if he chooses.

PR #164, opened 2021-01-02
Remove F_SETSIG

waywardmonkeys:
The first commit in this PR explains the details.


nbriggs:
The `fcntl` manual page is a little unclear about whether the act of setting the signal (even to the default) also enables the sending of any signals at all.  However, there is other (GNU) documentation on Interrupt driven I/O which says,

> If you set the O_ASYNC status flag on a file descriptor (see File Status Flags), a SIGIO signal is sent whenever input or output becomes possible on that file descriptor. The process or process group to receive the signal can be selected by using the F_SETOWN command to the fcntl function.

So it looks as though we can dispense with the F_SETSIG on Linux, , however we should make sure that the signal handler is properly set up, and that the O_ASYNC flag is set.

PR #166, opened 2021-01-03
Remove #ifdef I386 def for N_GETINUMBER.

waywardmonkeys:
There were 2 definitions for `N_GETINUMBER`, one for `I386` and one
for `I386` not being defined. Whether they differed or not in the
past isn't known, but they're identical now, so we don't need
two definitions.


blakemcbride (approved):

nbriggs (approved):
LGTM.

PR #169, opened 2021-01-03
Correctly terminate execl()

blakemcbride:


nbriggs (approved):
LGTM.

nbriggs (approved):
Doesn't fix a problem (some systems `#define NULL 0`), but it's harmless.

blakemcbride:
0 is 4 bytes on an X86 64-bit machine.  (char *) 0 is 8 bytes.  Since it is a variable argument function, this is definitely a problem.

nbriggs:
The 2005 ISO C standard says
>An integer constant expression with the value 0, or such an expression cast to type void *, is called a null pointer constant.[55]
> 55) The macro NULL is defined in <stddef.h> (and other headers) as a null pointer constant; see 7.17.

blakemcbride:
EXEC(3)                    Linux Programmer's Manual                   EXEC(3)

NAME
       execl, execlp, execle, execv, execvp, execvpe - execute a file

SYNOPSIS
       #include <unistd.h>

       extern char **environ;

       int execl(const char *path, const char *arg, ...
                       /* (char  *) NULL */);

PR #171, opened 2021-01-03
Pointer math correction

blakemcbride:


waywardmonkeys (changes requested):
I was working on a similar patch ... but in that, I'd changed the types of `len`, and `diff` and such to `size_t`. Would you mind doing the same here?

blakemcbride:
Done

waywardmonkeys (approved):
LGTM

nbriggs (approved):
LGTM

PR #173, opened 2021-01-04
Fix unused variable warnings.

waywardmonkeys:


nbriggs (commented):
The call to `DBPRINT()`, line 733, would use `terno`, but it is never set anywhere.
```
../src/unixcomm.c:695:13: warning: unused variable 'terno' [-Wunused-variable]
        int terno; /* holds errno thru sys calls after I/O fails */
            ^
```

waywardmonkeys:
That's why I didn't fix that one yet. It isn't just a matter of deleting it, but I wanted to look at fixing it.

nbriggs (approved):
Lets go with this as it is and pick up the other changes later.

PR #174, opened 2021-01-04
Fix warnings for unused variables when NOETHER is defined

nbriggs:
This doesn't fix unused variable warnings when NOETHER is NOT defined.


waywardmonkeys (approved):
LGTM

PR #175, opened 2021-01-04
Clean up some mixed sign issues

nbriggs:
LASTVERSIONARRAY, the marker value indicating the last entry, can be an int value as it is only compared to the int version_no field in a struct filename_entry.

MAXFINFO is the number of elements in the file info array -- doesn't need to be unsigned, and is used in signed contexts (it could be ssize_t, but we're not generally using that yet).


waywardmonkeys (approved):
LGTM

PR #177, opened 2021-01-04
Zero buffers before filling for write.

waywardmonkeys:
We were filling in parts of a buffer and writing it, but that
left some parts either uninitialized or with stale data from
the last read.

This clears out the buffer first so that we're not sending junk.


nbriggs (approved):
Yes.

PR #178, opened 2021-01-04
Pass correct sockaddr size to bind.

waywardmonkeys:
We should be passing the size of the `struct sockaddr_un`, not
the length of the path + the size of the path field.

There is another `bind` call in this file that had it right.


nbriggs (approved):
LGTM

PR #179, opened 2021-01-05
Change lisp_string_to_c_string length parameter to be size_t

nbriggs:
The argument passed to the length parameter of lisp_string_to_c_string is
always the sizeof an array, and since it is compared to the Lisp unsigned
length of the Lisp string is more appropriately a size_t than an int.
Likewise for the index into a string.


waywardmonkeys (approved):
LGTM. The Lisp strings will never be this big, but this is much better on the Maiko side.

nbriggs:
I was prompted to fix this by the "signed vs unsigned int comparison" warning.  I agree that the Lisp strings will never be this big.

PR #180, opened 2021-01-05
Fix tty.c to compile.

waywardmonkeys:
This doesn't output all of the same debug info, nor does it handle
setting the TTY into RAW mode yet.

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -9,7 +9,18 @@
 
 #include "version.h"
 
-#include <sys/select.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <sys/ioctl.h>
+#include <termios.h>
+#include <unistd.h>
+
+#include "lispemul.h"
+#include "lispmap.h"
+#include "adr68k.h"
+#include "lsptypes.h"
+#include "lspglob.h"
+#include "commondefs.h"
 #include "tty.h"
 
 DLTTY_OUT_COMMAND *DLTTYPortCmd;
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
If you know how to handle `RAW`, I'd love to hear. I'm not sure if the debug info that I didn't bring over is important.

nbriggs (commented):
"raw" for tty modes is documented as equivalent to:
```
stty cs8 -icanon min 1 time 0 -isig -xcase  -inpck -opost
```
Does that make sense?

waywardmonkeys:
I think it makes sense, thanks.

I'm working on some other changes for this as well.

waywardmonkeys:
Somehow, I'd previously missed `cfmakeraw()`. This is now ready for review and merge.

src/tty.c
```diff
@@ -152,83 +169,23 @@ void TTY_cmd() {
...
   if (TTY_Fd >= 0) {
-    stat = ioctl(TTY_Fd, TIOCGETP, &mode);
-    printf("DEBUG: \t\tTTY_Mode.sg_ispeed = %#x\n", mode.sg_ispeed);
-    printf("DEBUG: \t\tTTY_Mode.sg_ospeed = %#x\n", mode.sg_ospeed);
-    printf("DEBUG: \t\tTTY_Mode.sg_erase  = %#x\n", mode.sg_erase);
-    printf("DEBUG: \t\tTTY_Mode.sg_kill   = %#x\n", mode.sg_kill);
-    printf("DEBUG: \t\tTTY_Mode.sg_flags  = %#x\n", mode.sg_flags);
+    struct termios attr;
+    tcgetattr(TTY_Fd, &attr);
+    printf("DEBUG: \t\tTTY_Mode.sg_ispeed = %#x\n", cfgetispeed(&attr));
```
nbriggs, line 183: Couple of printf format warnings here because `speed_t` is `unsigned int` on FreeBSD (Linux?) but `unsigned long` on macOS.  Probably reasonable to cast to `unsigned int` since we're talking 0..38400.

src/tty.c
```diff
@@ -56,84 +60,97 @@ void tty_open() {
...
+{
+  speed_t baudrate;
 
-void tty_breakoff() {
-  int stat;
+  if (TTY_Fd >= 0) {
+    if ((baudrate = tty_baudtosymbol(DLTTYOut->line_speed)) != -1) {
+        struct termios options;
+        tcgetattr(TTY_Fd, &options);
+        cfsetispeed(&options, baudrate);
```
nbriggs, line 128: Nit: `cfsetspeed(&options, baudrate)` does both input and output speed at once, since we always make them the same.

src/tty.c
waywardmonkeys: Fixed and force pushed.

src/tty.c
waywardmonkeys: It ends up that `cfsetspeed` is a BSD 4.4 extension and not supported by Solaris. It also ends up that Solaris apparently doesn't have `cfmakeraw` either, so I've expanded that and force pushed.

nbriggs (approved):
Clearly I hadn't got to the Solaris box this AM :-(  Checked that now, and LGTM.

PR #181, opened 2021-01-05
Remove unused variables and cast return type correctly

nbriggs:
Remove unused variables and cast return value for `MakeAtom68k()` to suppress warning.


waywardmonkeys (approved):
LGTM

PR #182, opened 2021-01-05
Use empty initializer syntax to zero structs at declaration

nbriggs:
Fix "warning: missing field 'xxx' initializer" due to use of incomplete initializers in `xlspwin.c`.

### DIFF
--- a/src/xlspwin.c
+++ b/src/xlspwin.c
@@ -72,11 +72,11 @@ extern unsigned LispWindowRequestedWidth, LispWindowRequestedHeight;
 
 void Create_LispWindow(DspInterface dsp)
 {
-  XSizeHints szhint = {0};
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I'm in favor of this, but ... it is a GNU extension to C, right? Does it work in Solaris Studio?

waywardmonkeys:
Looking into this more, in C++, this is valid. In C, it is not and the Visual Studio compiler will error on it, I think.

The `= { 0 }` syntax should work though ... so I wonder why that is still issuing warnings?

nbriggs:
Yeah, no.  For Solaris.  I think we should just abandon this and chalk it up to compiler silliness.
It *does* generate the correct code to zero the structs with the `= {};` syntax with clang.

PR #184, opened 2021-01-05
Remove REGISTER define.

waywardmonkeys:
This was used to indicate that the value was allowed to live
in a register. Compilers are much better now and can handle
this themselves.

This helps get rid of more platform-specific ifdef-ery.


nbriggs (approved):
LGTM.  Also removes a few uses of lowercase `register` storage class also in floating point procedures.

PR #185, opened 2021-01-05
Remove I386-only, undefined macro usages: I386Round, I386Reset.

waywardmonkeys:
These macros aren't actually defined anywhere in our code and
aren't used outside of the old support for the Sun I386 hardware.

### DIFF
--- a/inc/my.h
+++ b/inc/my.h
@@ -56,8 +56,8 @@
 	}								\
 }
 #else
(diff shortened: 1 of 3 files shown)


nbriggs (approved):
LGTM.  The definition of `N_MakeFloat` in inc/my.h is now the same, except for the substitution of `LispPTR` for `int`, between the I386 and not I386 cases.  The fact that the I386 case didn't get updated from int to LispPTR is just a quirk of history.

PR #186, opened 2021-01-05
Remove redundant #ifndef NOETHER checks

nbriggs:
When the NOETHER checks were simplified in ldeether.c, there remained some
redundant #ifndef NOETHER within the #else of an #ifdef NOETHER.


waywardmonkeys (approved):
LGTM

PR #170, opened 2021-01-03
Don't close a file that hasn't been opened (2)

blakemcbride:

PR #187, opened 2021-01-06
Revert "Don't close a file that hasn't been opened (2)"

masinter:
Reverts Interlisp/maiko#170


masinter:
sorry, i mis-read the status

PR #188, opened 2021-01-06
Clean up warnings when compiling with ethernet enabled

nbriggs:
Add a dlpidefs.h header for prototypes.
Convert from bcopy() to memcpy().
Change some char to u_char as needed.


waywardmonkeys (approved):
LGTM

PR #189, opened 2021-01-07
Add include guards to all the include files in inc/ which were missing them

nbriggs:
As a prerequisite for cleaning up some other include issues, all the
include files in inc/ should have an include guard.  All the xxxdefs.h
were created with them, but most older files were not.


nbriggs (commented):
Looks as though I missed some -- the files that were named xxxdefs.h but were not ones that were created to provide prototypes for corresponding xxx.c file.

I'll push an update to this before it's ready to merge.

waywardmonkeys:
Not looking at this line by line as I am somewhat distracted by other things at the moment ... but I am in agreement with the thrust of this and think you should merge it.

nbriggs:
Fair enough.  I generated the list of files that needed this and the edits to implement it programmatically.
It enables me to then make sure that includes include what they use without duplicate problems.

PR #190, opened 2021-01-08
Migrate to POSIX signals

waywardmonkeys:


waywardmonkeys:
This is a WIP draft PR. It handles `sighold` and `sigrelse` so far.

A lot of checks are done for signals that are now standard, so `#if defined(SIGALRM)` and the like are no longer needed.

masinter:
Do we need signals at all? Maiko DSK doesn't use them. X mouse input and keyboard work by polling (on cygwin at least).  For networking, there's pretty much no one to talk to that we could and same with RS232.
so except for heartbeat signals...

maybe i don't understand what you mean by posix signals

nbriggs:
The timer is a signal.

waywardmonkeys:
I'd like to see a good bit of the signal code go away where possible. But, it'll be easier when there's something clean using the POSIX APIs instead of having 2 code paths in the codebase, one for SysV signals and one for BSD.

waywardmonkeys:
This is ready to go now.

nbriggs (changes requested):
I'm thinking that `int_panic_init()` should probably set up the handlers for all signals (sigfillset) then reset the ignores, and that `int_panic_init()` should be called before any other handlers are initialized in `int_init()`.

The default case in panicuraid could make use of `strsignal` to get a human readable name for the signal number (a few other cases could probably be collapsed with that too).

I'd rather that taking a signal like SIGUSR[12], or SIGXCPU not result in losing your workspace with:
>zsh: cpu limit exceeded  ldex test.vm
>

I am seeing some odd behavior on FreeBSD, but I haven't narrowed down whether it's new to this style of signal handling -- I'm guessing not, though.  Using truss on the ldex process from within a shell spawned from the ldex process I'm tracing can lock up, and truss'ing from an unrelated shell generates a SIGSUSP (I think) when you terminate the truss -- but it seems to recover mostly OK.

Otherwise, it's looking great.

waywardmonkeys:
I've pushed a new version with the changes to use `strsignal` now and collapsing some cases in `panicuraid`.

I don't want to modify the other logic too much right now as there are already some changes and I think we should look through the list of POSIX signals and see if we're actually doing what we should for each one, and that's handled better as a separate PR.

nbriggs (approved):
OK.  We certainly want to check that we're doing something appropriate for all POSIX signals, but the main point behind `panicuraid()` is that when some *unexpected* signal happens you get a chance to do something and possibly save your work before Lisp dies.  The current code removes any handling for SIGABRT, SIGLOST, SIGUSR1, SIGUSR2, and SIGXCPU

nbriggs:
BTW -- the Solaris SPARC system is unavailable at the moment so I haven't compiled this branch there yet.  I'll try on an X86_64 Solaris in a short while.

waywardmonkeys:
`SIGABRT`, `SIGLOST`, and `SIGXCPU` were only handled on SunOS4, which is no longer a supported OS. `SIGUSR1` and `SIGUSR2` were only handled in the BSD code path, not via `SYSVSIGNALS`. I could add them here too.

seems like a solid audit would be a good thing.

nbriggs (approved):
OK, let's merge and review the signals that we catch.

PR #191, opened 2021-01-08
Remove old (NEVER) defs for int_timer_on|off.

waywardmonkeys:
These were replaced by other calls that manage more signals
than just the `SIGVTALRM`.


nbriggs (approved):
LGTM

PR #192, opened 2021-01-08
ptsname() needs _GNU_SOURCE on glibc.

waywardmonkeys:
glibc needs `_XOPEN_SOURCE` to have a declaration for
`ptsname()`. An easy way to get that without having to
know about particular `_XOPEN_SOURCE` values is to define
`_GNU_SOURCE`.

Closes interlisp/medley#137.


nbriggs (approved):
LGTM.

PR #195, opened 2021-01-08
Remove OCR code.

waywardmonkeys:
This is very dated code that assumes there's a `/dev/ocr0`.

I'm not sure what system this was for, but it doesn't appear to
be one that exists currently.

Discussed in interlisp/medley#126.


waywardmonkeys:
There's a lot of very old signal handling code in this file among other things, so I keep seeing it when grepping.

nbriggs (approved):
LGTM.

nbriggs:
I'm guessing this code hooked up to a SCSI connected OCR scanner, probably a Japanese office device.
The code isn't remotely interesting today except as a reference for how one might add call-outs to slow external devices.
I think the code can be deleted.

nbriggs:
Hi... @masinter you added yourself as a reviewer, did you want to add something to the review comments?

masinter (approved):
sorry, i was looking for a way to indicate approval

PR #198, opened 2021-01-09
Experimental build support via cmake.

waywardmonkeys:
This doesn't remove or replace the existing build system, but
provides an alternative. It allows for faster builds with proper
incremental builds, as well as using tools such as `clang-tidy`,
which work better when they have a compiler commands 'database'.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,476 @@
+PROJECT(maiko C)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.15)
+
+SET(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+SET(CMAKE_C_STANDARD 90)
+
+IF(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_COMPILER_ID MATCHES "GNU")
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
To use this, you can:

```
mkdir build
cd build
cmake ..
make
```

If you want to use `ninja` with it, you would:

```
```
mkdir build
cd build
cmake .. -GNinjja
ninjja
```

There's little bits of pieces to do, like making it work on Solaris, adding a configuration option for X11 stuff, making it work on Cygwin, etc.

This isn't (yet) a proposal that we switch to this all the time. But I end up having this in almost every branch that I work on for driving basic / regular development.

nbriggs:
I ran cmake on Solaris when you first made a CMakeList.txt available -- no trouble there.  I'll try this version when I have access to the SPARC system again (tomorrow, if the ISP/network issue gets resolved)

nbriggs (approved):
LGTM, presumably it will need update with the removal of oether.c and oldeether.c

masinter:
should the cmake instructions be added to the README?

waywardmonkeys:
I'll submit a PR for that this weekend.

PR #202, opened 2021-01-10
Remove extra flags for gcc on some platforms.

waywardmonkeys:
On Linux x86 and x86_64 as well as Cygwin, we were adding some
flags to the (commented out) usage of gcc. These shouldn't be
here.


nbriggs (approved):
LGTM.  These were for debugging, but I *think* that with current versions of gcc the `-fno-agressive-loop-optimizations` is unnecessary -- and these were commented out so it clearly won't affect the builds unless people were switching.

PR #203, opened 2021-01-10
Switch to compiling as ISO C99 + GNU extensions.

waywardmonkeys:
We currently build for ISO C90 with GNU extensions. The GNU
extensions include both library and language features.
Compiling without GNU extensions results in a number of compilation
failures due to use of those library features.

With this, we'll build for ISO C99 with the GNU extensions and
can evaluate what's going on with those library features in the
future.


waywardmonkeys:
I feel like the `bin/makefile-sunos5.sparc-x` may need to be updated if it is calling the Solaris Studio tools.

waywardmonkeys:
Differences in the language versions can be found in http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf (page 13) where section 6 is changes from C99 to C11 and section 7 is C89 to C99.

There's a reasonably solid argument that we could probably just move to C11 instead as well.

nbriggs:
Still waiting on Solaris SPARC access.  I will check on the Solaris 11.4 x86_64 with the Solaris Studio compilers.   The SPARC box currently has Solaris Studio, clang (11), and gcc (10+ snapshot from git repo)  all installed with 32- and 64-bit multi-arch support.

nbriggs (approved):
I'm going to go ahead and approve this as is, we can fix any Solaris problems that show up when I can test it.

nbriggs:
The Solaris on SPARC system is back.  The developer studio compiler accepts the "-std=c99" as the other compilers do, so we can update the `bin/makefile-sunos5.sparc-x` similarly to the others.

PR #204, opened 2021-01-10
Remove I386 version of N_MakeFloat.

waywardmonkeys:
This used to differ, but the differences were dead code and
removed. We have no need for this extra bit of baggage.


nbriggs (approved):
LGTM

PR #205, opened 2021-01-10
Remove SunOS 4 version of aref_switch.

waywardmonkeys:
When compiling for SunOS4, we would have a macro-ized version
of this function which required a slightly different call-site.

This removes all of that as it isn't used or needed.


nbriggs (approved):
LGTM.

PR #206, opened 2021-01-10
MDate: Make time_t, not long.

waywardmonkeys:
MDate, generated into `vdate.c` from `mkvdate.c`, was being
stored as a `long` rather than a `time_t`. This led to some
casts, but also a bit of platform #ifdef'd code.

This makes that go away by storing it as the `time_t` value
that it really is.


src/mkvdate.c
```diff
@@ -52,7 +52,8 @@ int main(void) {
   fprintf(stderr, "Mdate :%ld\n", (long)time.tv_sec);
   fprintf(stderr, "Version: %s\n", ctime(&time.tv_sec));
 
-  printf("long MDate= %ld;\n", (long)time.tv_sec);
+  printf("#include <time.h>\n");
+  printf("time_t MDate= %ld;\n", time.tv_sec);
```
nbriggs, line 56: ../src/mkvdate.c:56:34: warning: format specifies type 'long' but the argument has type 'time_t' (aka 'int') [-Wformat]
  printf("time_t MDate= %ld;\n", time.tv_sec);
                        ~~~      ^~~~~~~~~~~
                        %d

waywardmonkeys: Ugh. I updated with a force push that should address that.

nbriggs (approved):
LGTM.  Approved as is but I'll wait a moment in case you want to change the comment at the front (mkvdate.c:23) and/or the mixture of " :" and ": " (mkvdate.c:57) in the output while you're in this file.

waywardmonkeys:
Okay, re-pushed with various nits fixed.

PR #208, opened 2021-01-12
Switch `s_char` uses to `signed char`. Remove s_char alias.

waywardmonkeys:


waywardmonkeys:
This brings us a bit further along the path to having few to no platform ifdefs.

nbriggs (commented):
Do we have `int8_t` at this point?  That's what they should really be since they're small integer offsets from the (Lisp) PC.

waywardmonkeys:
We do. Let's see how that goes.

waywardmonkeys:
I sort of want to have inline functions for reading various types from the Lisp world.

```
int8_t maiko_read_int8(ADDR);
uint8_t maiko_read_uint8(ADDR);
```
and so on ... this would let us go through and replace most / all `GETBYTE` calls with something that indicates the types involved. Inlining would remove all overhead.

But that would be a separate PR. This at least lets us get rid of `s_char`.

waywardmonkeys:
(and this has been updated now to use `int8_t`.)

nbriggs:
Doesn't help with `GETBYTE(((char *)Addr68k_from_LADDR(base)) + (index >> 3)) |= new`, or `GETWORD(CurrentStackPTR + 2) = savestack1` which is used fairly extensively.

nbriggs (approved):
LGTM.

nbriggs:
BTW -- [Source to source translation](https://patents.google.com/patent/CA2201882A1), maintaining the "puns" in macros etc. was work that my wife was responsible for back in 1994.  She was translating Tandem PTAL to C, producing readable C that the author of the PTAL could then maintain directly.  Much nicer than the "C is our assembly language" approach that other systems around that time were doing.

PR #209, opened 2021-01-12
Default NOFORN to defined.

waywardmonkeys:
The NOFORN flag controls foreign function interface code, which
was written using the dld_* APIs which haven't existed in years.

There was a GNU dld that provided this API, but it was withdrawn
in 2006.

If this code is to be re-enabled, it will have to be changed to
use the `dlopen` family of APIs.


waywardmonkeys:
This takes us another step towards having less config-that-is-default exposed in the build system for every platform.

nbriggs (approved):
LGTM

PR #210, opened 2021-01-12
Clean up CMakeLists.txt for Solaris (SunOS5) builds with ethernet 

nbriggs:
dlpidefs.h was missing from the list of headers.
Solaris compiles should have USE_DLPI set.
ldeether depends on the dlpi object in the maiko library.


waywardmonkeys (approved):
LGTM

PR #211, opened 2021-01-12
Fix lispcmp bytecmp

nbriggs:
Cleans up some type mismatches and makes local helper functions static and remove them from the header file.

### DIFF
--- a/inc/mkatomdefs.h
+++ b/inc/mkatomdefs.h
@@ -5,7 +5,7 @@ DLword compute_lisp_hash(const char *char_base, DLword offset, DLword length, DL
 LispPTR compare_chars(register const char *char1, register const char *char2, register DLword length);
 int bytecmp(const char *char1, const char *char2, int len);
 LispPTR compare_lisp_chars(register const char *char1, register const char *char2, register DLword length, DLword fat1, DLword fat2);
-int lispcmp(const DLword *char1, const unsigned char *char2, int len);
+int lispcmp(const DLword *char1, const char *char2, int len);
 LispPTR make_atom(const char *char_base, DLword offset, DLword length, short int non_numericp);
 LispPTR parse_number(const char *char_base, short int length);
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
If feels odd to me that we're saying here that we're dealing with `unsigned char`, but then the prototypes for the functions are for `char`. Is this a place where we should be using `uint8_t` instead and being more clear?

nbriggs:
I thought about that, and it added some casts -- it looks like:
```
int lispcmp(const DLword *char1, const uint8_t *char2, int len)
```
and calls with
```
   if (lispcmp((DLword *)char1, (uint8_t *)char2, length))
   if (lispcmp((DLword *)char2, (uint8_t *)char1, length))
```
and it's going to apply `GETWORD` and `GETBYTE` to the first and second argument respectively.

For bytecmp it's better --
```
int bytecmp(const uint8_t *char1, const uint8_t *char2, int len)
      if (GETBYTE(char1++) != *char2++) return (0);
```
where char1 is a pointer to a Lisp possibly byte-swapped string and char2 is a native host string, at least in the caller's caller.
Remember that GETWORD and GETBYTE are
```
#define GETWORD(base) (* (base))
#define GETBYTE(base) (* (base))
```
or when little-endian
```
#define GETWORD(base) (* (DLword *) (2^(UNSIGNED)(base)))
#define GETBYTE(base) (* (unsigned char *) (3^(UNSIGNED)(base)))
```
and we could change that `unsigned char` in `GETBYTE` to a `uint8_t` as well.

It would be nice if the code had made better use of the type system to track what's a pointer that needs adjustment and what's a native pointer, but I gather it interfered with what little optimization they could get.

waywardmonkeys (approved):
LGTM

nbriggs:
I'll merge this, but keep it in mind for a bit more cleanup at some other time.

PR #212, opened 2021-01-12
Merge hacks.c into foreign.c.

waywardmonkeys:
This code was not really valid ANSI C in the way that it was
used, and it was only used by foreign.c, which is #ifdef'd out.


nbriggs (approved):
LGTM

PR #215, opened 2021-01-12
Remove #ifdef LINUX max/min definitions.

waywardmonkeys:
These don't appear to be needed.


nbriggs:
I was just about to ask about that w.r.t. #213

nbriggs (approved):
LGTM.  Checked with clang 10.0 on Ubuntu 20.04.

PR #216, opened 2021-01-12
Add maiko/platform.h to detect OS + CPU.

waywardmonkeys:
This also adds a bit to the `-info` output to say which OS and CPU
are being targeted.

This can be used in subsequent commits to drive whether or not
we need `BYTESWAP` and whether or not unaligned memory access
is okay.

We'll be able to remove per-platform defines from all of the
makefiles and cmake.


nbriggs (commented):
I'll look and see what my PowerBook G4 admits to.  It also doesn't tell us whether we compiled 32-bit or 64-bit -- I do both on my MacBook because it's still running El Capitan that can do either.  The SPARC box also does both, but different compilers have different defaults if you don't say.

nbriggs:
Here's the output for the C compiler defined macros on my OS X 10.5.8 Powerbook G4.
[osx-10.5.8-ppc.txt](https://github.com/Interlisp/maiko/files/5805441/osx-10.5.8-ppc.txt)

Since I managed to get that machine back on the network and get git access from it to work, I tried compiling there and everything seemed OK.  I'll check in a makefile-darwin.ppc-x.  There's no cmake on that system, and I don't want to try building it from source there.

nbriggs (approved):
LGTM.  The OS detection works for Mac OS X 10.5 on PowerPC as is.  The ARCH detection can use defined(__ppc__)  || defined(_ARCH_PPC).  Not sure if we want the architecture name to be PPC (or ppc) or PowerPC.

waywardmonkeys:
I've added more.

It now detects x86_64 under Oracle Studio better, RISC-V, PowerPC, and adds a define for 32 vs 64 bits for each target architecture.

nbriggs:
Even better.  Most (but not all) of the compilers I've looked at set __BYTE_ORDER__ to one of __ORDER_LITTLE_ENDIAN__, __ORDER_BIG_ENDIAN__, or __ORDER_PDP_ENDIAN__, and then there's a mix of whether they set __{LITTLE|BIG}_ENDIAN__  (clang, but not gcc or developer studio).

Does anyone maintain a web collection of the defines that the compilers have on various systems?

waywardmonkeys:
There is Boost predef, an older predef project on sourceforge... but they are sometimes incomplete! And they don’t just show what defines are available but do something similar to my header here.

nbriggs:
On Solaris 11.4, gcc can't detect the OS because it defines `__SVR4` but not `__SVR4__`

waywardmonkeys:
I'll fix that then.

waywardmonkeys:
Force pushed with a fix for that. I also defined the macros with a value of `1` rather than nothing as well.

As for endian ...there's existing headers like:

https://www.boost.org/doc/libs/1_54_0/boost/detail/endian.hpp
https://opensource.apple.com/source/WTF/WTF-7601.1.46.42/wtf/Platform.h.auto.html

For the most part, we can base it on the CPU arch ... but we will have to deal with `__ppcle__` and `__ppc64le__` and `__ARMEB__` and so on ...

waywardmonkeys:
Onward and upward!

PR #217, opened 2021-01-13
Rename NOFORN to MAIKO_ENABLE_FOREIGN_FUNCTION_INTERFACE.

waywardmonkeys:
This lets us get rid of double-negatives, which are confusing,
and starts a naming scheme that is easier to see when something
is a Maiko flag versus something else.


nbriggs (approved):
I agree.  I might have chosen FFI rather than spelling it out, but I'm not wedded to that.

Presumably we'll be going with MAIKO_ENABLE_ETHERNET in another PR.

PR #218, opened 2021-01-13
Remove HAS_GETHOSTID.

waywardmonkeys:
This is only true on DOS now. It was previously true on older
unsupported platforms, but they are gone now.


nbriggs (approved):
LGTM.

PR #219, opened 2021-01-13
Rename NOETHER to MAIKO_ENABLE_ETHERNET.

waywardmonkeys:
Like the renaming of NOFORN, this gets rid of some double
negatives. It also removes some patterns where we had:

```
    #ifdef NOETHER
    #else
    ...
    #endif
```

and replaces them with:

```
    #ifdef MAIKO_ENABLE_ETHERNET
    ...
    #endif
```

### DIFF
--- a/inc/version.h
+++ b/inc/version.h
@@ -206,7 +206,6 @@ error Must specify RELEASE to build Medley.
 
 
 	/* Set up defaults */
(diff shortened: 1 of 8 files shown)


nbriggs (approved):
LGTM

PR #220, opened 2021-01-13
Use sig_atomic_t for FP_error when using FP exceptions.

waywardmonkeys:
`sig_atomic_t` is usually `int`, but is the typedef that one
should use for flags written to from a signal handler.


nbriggs (approved):
LGTM.

PR #221, opened 2021-01-13
Remove unused FileIOFlag, TimerFlag.

waywardmonkeys:


nbriggs (approved):
This seems OK.  Equivalent information seems to be maintained in the bits in the Interrupt State word.  We need to be sure that the bits there do get updated based on (interrupt) events.

PR #223, opened 2021-01-14
unixfork.c needs to define `flushing`.

waywardmonkeys:
This shouldn't be an `extern` decl here as `unixfork.c` is built
into `lde` which doesn't get a copy of `main.c` linked in.

### DIFF
--- a/src/unixfork.c
+++ b/src/unixfork.c
@@ -42,7 +42,7 @@
 
 #ifdef DEBUG
 /* required by DBPRINT from dbprint.h */
-extern int flushing = 0;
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM.  What surfaced that one?  It shouldn't have shown up as a compiler warning as it's both declaration and definition with the initializer present.

waywardmonkeys:
I have patches that add a config flag to cmake to turn on `DEBUG` and that generated a warning. There are a bunch of other warnings too, but that one was unique.

PR #224, opened 2021-01-14
GitHub Actions: Only run build-pi for Interlisp/maiko

waywardmonkeys:
The build-pi builder is self-hosted, so it may not be present,
and isn't present in forks of this repo, making the CI less
useful.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,7 @@ jobs:
 
   build-pi:
     runs-on: [self-hosted, linux, ARM]
+    if: ${{ github.repository == 'Interlisp/maiko' }}
     steps:
       - uses: actions/checkout@v2
       - name: Build


nbriggs:
You know infinitely more about this area than I do.  I think this is running with Larry's RPi, so I'll let him approve and merge unless he punts it back to me.

waywardmonkeys:
Well, until then, every push that I do to my fork fails the GitHub Actions hooks. :)

nbriggs:
OK.  I believe it's OK, so I'll merge it now.

masinter:
I'm sorry I pushed this to master. I was exploring with @AbeJellinek showing my self-hosted runner.
Thanks for cleaning up the mess I left. I thought I set up that I can't push to master without a review but I guess that's only in medley repo.

PR #225, opened 2021-01-14
GitHub Actions: Update syntax for installing a cask from homebrew.

waywardmonkeys:
Homebrew on the macOS build machines for GitHub Actions was
updated to version 2.7 and that deprecated the `brew cask install`
syntax in favor of `brew install --cask`, so we need to update
here.

The announcement for the GitHub change is:

https://github.com/actions/virtual-environments/issues/2415


nbriggs (approved):
LGTM.

PR #227, opened 2021-01-15
Correct some ifdefs from ISC to RESWAPPEDCODESTREAM.

waywardmonkeys:
`RESWAPPEDCODESTREAM` was only used on ISC, so some `#ifdef`'s were
previously set up incorrectly.

### DIFF
--- a/inc/return.h
+++ b/inc/return.h
@@ -33,7 +33,7 @@
 
 
 /* FAST case return use */
-#ifndef ISC
+#ifndef RESWAPPEDCODESTREAM
 #define FastRetCALL							\
   {									\
     /* Get IVar from Returnee's IVAR offset slot(BF) */ 			\
@@ -60,7 +60,7 @@
 	FuncObj->byteswapped = 1;				\
       }								\
   }
-#endif /* ISC */
+#endif /* RESWAPPEDCODESTREAM */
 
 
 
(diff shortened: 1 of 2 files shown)


inc/return.h
```diff
@@ -60,7 +60,7 @@
 	FuncObj->byteswapped = 1;				\
```
nbriggs, line 60: This code fragment also shows up in tosfns.h, SWAPPED_FN_CHECK expanding to this if RESWAPPEDCODESTREAM, or nothing if not.  Not sure if it's worth shrinking the FastRetCALL down here.

waywardmonkeys: If this code is made to work again ... I suspect we'd be able to simplify a bunch of it. But even when I did this, things didn't compile afterwards and then I got sick for a day and I forgot where I was at with it.

src/mvs.c
```diff
@@ -72,12 +72,12 @@ LispPTR values(int arg_count, register LispPTR *args) {
     fnhead = (struct fnhead *)Addr68k_from_LADDR(POINTERMASK & SWA_FNHEAD((int)caller->fnheader));
     pc = (ByteCode *)fnhead + (caller->pc);
   }
-#ifdef ISC
+#ifdef RESWAPPEDCODESTREAM
```
nbriggs, line 75: Same thing with SWAPPED_FN_CHECK, except it would have to be rewritten to take as a parameter the name of the function object (fnhead here, FuncObj above).

nbriggs (approved):
This seems correct for the changes as proposed.  See other comments attached to specific lines.

PR #228, opened 2021-01-15
Remove incorrect SUN3_OS3_OS4_il ifdef.

waywardmonkeys:
The definition of a function was ifdef'd out, making it seem
like perhaps there was an assembly implementation, but there
isn't. That platform support is dead weight also at this point,
so removing this because it isn't something that will come back
is fine.


nbriggs (approved):
LGTM.

PR #229, opened 2021-01-16
Remove -funsigned-char from standard compilation options

nbriggs:
Using -funsigned-char causes the Oracle Developer Studio C compiler typedefs for
int8_t to become unsigned (uint8_t) rather than signed, with no warning or
other indication that this has happened.

Add standardized options for Oracle Developer Studio C compiler (aka SunPro C)

Fixes Interlisp/medley#152


waywardmonkeys (approved):
LGTM. Thanks for picking up the "SunPro" change in cmake!

PR #230, opened 2021-01-16
Remove unnecessary codepaths for ISC.

waywardmonkeys:
ISC support had some extra code paths that were largely
duplicates of what was for non-ISC. They are not needed
any longer.


nbriggs (approved):
LGTM

PR #232, opened 2021-01-17
Delete code related to NATIVETRAN feature.

waywardmonkeys:
This was obsolete work that had been done for generating
native code from the bytecode.

Closes Interlisp/medley#89.


waywardmonkeys:
Trying to clear out some of the ifdefery in the dispatch and related code ... slimming it down is making it easier to understand.

src/testtool.c
```diff
@@ -1196,7 +1196,6 @@ void all_stack_dump(DLword start, DLword end, DLword silent)
           print(get_framename((struct frameex1 *)stkptr));
           printf(" [");
           if (((FX *)stkptr)->fast) printf("fast,");
-          if (((FX *)stkptr)->native) printf("native,");
```
nbriggs, line 1199: This one should probably stay until we remove the "native" flag bit from the Frame Extension, likewise for the Function Header struct which also has a "native" flag.

waywardmonkeys: I removed them in `inc/stack.h` and marked what they used to be. Or are you talking about Lisp side? But C side, everything is gone and a few foot prints left as a trail to the past.

nbriggs: Looks as though I got started in the middle somewhere, so I'd missed that you'd dummied out the "native" flag bit.
If all the traces (bar history pointers) are gone, then this is fine.  I/we do need to look at the Lisp side to make sure there are no dangling references into the emulator code.

nbriggs: Looks as though I got started in the middle somewhere, so I'd missed that you'd dummied out the "native" flag bit.
If all the traces (bar history pointers) are gone, then this is fine.  I/we do need to look at the Lisp side to make sure there are no dangling references into the emulator code.

nbriggs:
This will take me a little while to review -- I'll look at it in the morning so I have uninterrupted time.

nbriggs (commented):
LGTM (Remove unused SaveD6)

waywardmonkeys:
Anything pending on this?

nbriggs:
Not yet.  I'll be on it soon though.

nbriggs (approved):
LGTM.

PR #233, opened 2021-01-18
Remove old xXWINDOW code.

waywardmonkeys:
This looks like code that was a mix of SUNDISPLAY and XWINDOW.
But it isn't valid any longer...


nbriggs (approved):
LGTM

PR #234, opened 2021-01-18
Fix !XWINDOW to be SUNDISPLAY ifdefery.

waywardmonkeys:
Some of the `ifndef XWINDOW` code should have been
`ifdef SUNDISPLAY` instead.


nbriggs (approved):
LGTM

PR #235, opened 2021-01-18
Remove obsolete testdsp.

waywardmonkeys:
This was an old executable from the SUNDISPLAY days and is not
needed any longer.


waywardmonkeys:
Force pushed to resolve conflicts.

nbriggs:
OK.  If we want to test displays like this again we can always check back in the git history.

PR #236, opened 2021-01-18
Fix sunwindow include ifdefery on vmemsave.c.

waywardmonkeys:
The cursor management code is `#ifdef` `SUNDISPLAY` and
`OLD_CURSOR`, not `NOPIXRECT`, so do the same for the
header include.


nbriggs (approved):
LGTM

PR #237, opened 2021-01-19
Bit of cleanup for Lisp_Xinitialized.

waywardmonkeys:
* Remove the `extern` from `xcursor.c`, where it was not used.
* Move the definition from `main.c` to `xinit.c` as it is
  the only file using it.
* Use `stdbool.h` for it rather than an `int` with custom `TRUE`
  and `FALSE` values.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -249,7 +249,6 @@ char keystring[128] = {""};
 #define FALSE 0
 #define TRUE !FALSE
 
-int Lisp_Xinitialized = FALSE;
 char sysout_name[MAXPATHLEN]; /* Set by read_Xoption, in the X version. */
 int sysout_size = 0;    /* ditto */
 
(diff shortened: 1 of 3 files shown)


nbriggs (commented):
Lisp_Xinitialized looks as though it's a write-only variable...
```
% grep -r Lisp_Xinitial src inc include
src/main.c:int Lisp_Xinitialized = FALSE;
src/xcursor.c:extern int Lisp_Xinitialized, Bitmap_Pad, Default_Depth;
src/xinit.c:extern int Lisp_Xinitialized;
src/xinit.c:  Lisp_Xinitialized = FALSE;
src/xinit.c:  Lisp_Xinitialized = TRUE;
```
so it's unclear what it's bringing to the game.

waywardmonkeys:
Figured I might add some asserts to make sure things are set up and torn down correctly.

nbriggs (approved):
LGTM

PR #238, opened 2021-01-19
Remove FORKCOMM define.

waywardmonkeys:
This code path was followed everywhere except DOS, so remove
it from the build system, and adjust the code.


nbriggs (approved):
LGTM

PR #239, opened 2021-01-19
Change icon, cursor images and masks to const uint8_t

nbriggs:
The predefined cursors (image and mask) are better represented
as arrays of const uint8_t rather than char.  Likewise for the
window icon.  Track this change in the functions that take the
image and mask as arguments, casting only when we get to the
X library functions that take char*.

### DIFF
--- a/inc/xbitmaps.h
+++ b/inc/xbitmaps.h
@@ -11,40 +11,40 @@
 /*									*/
 /************************************************************************/
 
-
+#include <stdint.h>
 
 
 
 
 typedef struct {
 	short cubitsprepixel;	/* bits per pixel in the cursor, mostly 1 */
-	char *cuimage;		/* the image bitmap */
-	char *cumask;		/* the mask bitmap */
+	const uint8_t *cuimage;	/* the image bitmap */
+	const uint8_t *cumask;	/* the mask bitmap */
 	short cuhotspotx;	/* hot-spot X coordinate */
 	short cuhotspoty;	/* hot-spot Y coordinate */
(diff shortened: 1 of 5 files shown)


inc/xbitmaps.h
```diff
@@ -339,7 +339,7 @@ static char plain_bits[] =  {   0x00, 0x00,
 int Lisp_icon_width = 88;
 int Lisp_icon_height = 58;
 
-char Lisp_icon[] = 
+const uint8_t Lisp_icon[] =
```
waywardmonkeys, line 342: There's some extra whitespace at the end of this line that we might as well trim.

inc/xbitmaps.h
```diff
@@ -11,21 +11,21 @@
...
 
 
 
 
 typedef struct {
 	short cubitsprepixel;	/* bits per pixel in the cursor, mostly 1 */
-	char *cuimage;		/* the image bitmap */
-	char *cumask;		/* the mask bitmap */
+	const uint8_t *cuimage;		/* the image bitmap */
+	const uint8_t *cumask;		/* the mask bitmap */
```
waywardmonkeys, line 22: Can these comments be kept lined up?

waywardmonkeys (approved):
LGTM with a couple of nits.

inc/xbitmaps.h
nbriggs: Yes.

waywardmonkeys (approved):
LGTM

PR #240, opened 2021-01-19
Remove dead stores.

waywardmonkeys:


nbriggs (commented):
Some of these are dead and should go away, and some of them are dead but very handy when you're in the debugger trying to figure out what happened one or two cycles back.  With the compiler now almost certainly eliminating dead stores that's not as helpful as it would have been.  Not sure what the debugging strategy should be in those places.

waywardmonkeys:
I'm hopeful that by converting some of the macros into C99-style inline functions, we'll be able to make debugging a bit easier by having more capabilities accessible from the debugger. But all of these here either trigger warnings in some configuration or clang-tidy issues.

nbriggs (approved):
LGTM.

PR #242, opened 2021-01-19
conspage: Fix uninitialized memory read.

waywardmonkeys:
If `prior` was not set, then `priorpg` would be uninitialized and
was subsequently used in a comparison. It should be initialized
to `NULL`.


nbriggs (approved):
LGTM.  This looks like a clang-tidy suggestion.

waywardmonkeys:
Yes. We are getting pretty tidy.

nbriggs:
clang analyzer version 6.0 says 99 issues, version 11.0 finds 169 issues.

waywardmonkeys:
Used to be worse. :) my local branches are better than `master`

PR #243, opened 2021-01-19
Use the C99 floating point error code on Solaris.

waywardmonkeys:
Previously, we were building as C89 and the new code required
the C99 flag on Solaris with Sun Studio. Now that we build as
C99, this should work now and we can remove the special case
code.


nbriggs (approved):
LGTM.  Checked with Studio C on Solaris 11.4 SPARC.

PR #245, opened 2021-01-20
Move SunOS 3 and 4 makefiles to legacy directory.

waywardmonkeys:
We don't remove them for now as we still reference the build
options as we work through things.


nbriggs (approved):
LGTM.

PR #246, opened 2021-01-20
Remove csh version of makeright.

waywardmonkeys:
In 2001, makeright was ported from csh to bash. This is the old
csh version.


nbriggs (approved):
LGTM

PR #247, opened 2021-01-20
Remove macOS 11 from GitHub Actions.

waywardmonkeys:
This platform is experimental on their end and is having
capacity problems. We don't need it for now and the noise
of the failures is not worth it.


nbriggs (approved):
Yes.

PR #248, opened 2021-01-20
Remove endiffix relics.

waywardmonkeys:
This isn't something we need around or even have the full source
for. It is a SPARC binary and a lexer script.


nbriggs (approved):
LGTM

PR #249, opened 2021-01-20
Remove bin/launch.asm.

waywardmonkeys:
This was a copy of `src/launch.asm`. Part of the DOS build process
made a copy, but that isn't something that needs to be in version
control.


nbriggs (approved):
Agreed.  LGTM.

PR #250, opened 2021-01-20
Remove bin/runlisp*

waywardmonkeys:
These are from when people were running lde on a small set of hosts
and needed some keys hardcoded. They are no longer relevant.


nbriggs (approved):
Agreed.

PR #251, opened 2021-01-20
Remove key checking code.

waywardmonkeys:
This wasn't actually hooked up any more, so it was pretty much
all dead code. The handling of this in the command line options
is gone.

In `bin/makefile-tail`, `DEVFILES` and `LIBFILES` became identical, so
`DEVFILES` went away.


nbriggs (approved):
LGTM.  We haven't built with key checking in a long long time.

PR #252, opened 2021-01-20
Use MAIKO_ARCH_SPARC rather than `sparc`.

waywardmonkeys:


waywardmonkeys:
In a future commit / PR, I want to remove the `#ifdef LINUX` bit that undefs `UNALIGNED_FETCH_OK` since that's not OS specific but rather specific to the CPUs involved. At that point, we'll want to think about which other architectures should be flagged here. Maybe `MAIKO_ARCH_ARM`?

nbriggs (approved):
LGTM.

nbriggs:
I think older ARMs don't do unaligned access but ARM64s do -- but doesn't C impose a restriction on pointer alignment beyond the processor capabilities?

waywardmonkeys:
@nbriggs Yes ... but that isn't an issue here (the C restrictions). Here, we're just using this flag in one place:

```
/***** Get_DLword(ptr) ptr is char* ***/
#ifndef UNALIGNED_FETCH_OK
#define Get_DLword(ptr)     ((Get_BYTE(ptr) <<8) | Get_BYTE(ptr+1))
#else
#define Get_DLword(ptr)     *(((DLword *)WORDPTR(ptr)))
#endif
```

We'd probably even be fine to just drop the `DLword` cast and always do 2 `Get_BYTE`s.

PR #254, opened 2021-01-20
xc.c: Remove sparc-only optimization hack.

waywardmonkeys:
It looks like this was trying to force a pointer into a register,
but only for SPARC.


nbriggs (approved):
The (Lisp) machine state is a hot item holding the PC etc., so it's not a surprise that it helped to force it into a register.

PR #257, opened 2021-01-20
Remove ANSICC and related variables.

waywardmonkeys:
In the SunOS 4 days, the system cc apparently couldn't compile
the lp* files, so there was an arrangement to use an `ANSICC`,
which was just `gcc`, to build those files.

We no longer have this issue.


nbriggs (approved):
LGTM.  It was K&R vs ANSI.

PR #258, opened 2021-01-21
Disallow UNALIGNED_FETCH_OK on ARM, not LINUX.

waywardmonkeys:
UNALIGNED_FETCH_OK shouldn't be set for an entire OS, but
based on the CPU architecture.

Some ARM systems don't allow unaligned reads. x86 and x86_64
do, so update how we configure this.


nbriggs (approved):
Looks better.

PR #259, opened 2021-01-21
move Build instructions to wiki

masinter:


waywardmonkeys (approved):
I'm really not a fan of using a wiki for these sorts of things. But I guess that I'm okay with these changes, assuming that we're going to go with the wiki.

masinter:
As with Medley, build instructions are here in README but additional context from medley-intro is still in wiki.
Reviewers please review the wiki too.

PR #261, opened 2021-01-21
Remove flags for LINUX, OPENBSD, FREEBSD, MACOSX from build system.

waywardmonkeys:
We now can handle these via `inc/maiko/platform.h` and the
constants defined there.

This doesn't change `OS5` for Solaris yet as that's a much wider
set of changes.


nbriggs (approved):
Works for me.  Though my testing was complicated trying with -DSTACKCHECK  (Interlisp/medley#162)

PR #262, opened 2021-01-21
Use the same TIMER_INTERVAL on all platforms.

waywardmonkeys:
This removes a special case for SPARC which was using a slower
interval.


nbriggs (approved):
LGTM.  Tested on SPARC

PR #264, opened 2021-01-21
Remove find-writes and fixid.

waywardmonkeys:


nbriggs (approved):
LGTM.  These are no longer of any interest.

PR #265, opened 2021-01-21
Remove 'release' support from makeright.

waywardmonkeys:
This is the old way that releases were built and isn't relevant
today. It made a variety of assumptions about how things were
linked and allowed for the user to re-link an executable, but
that isn't something we need to do any longer.

This doesn't yet remove the checksum, ldechecksum, and related
scripts.


nbriggs (approved):
LGTM.

PR #267, opened 2021-01-21
cmake: ldeether shouldn't link full maiko lib.

waywardmonkeys:
In the old build system, `ldeether` is built as follows:

```
$(OSARCHDIR)ldeether :  $(SRCDIR)ldeether.c $(DLPIFILES)
    $(CC) $(CFLAGS) -I$(INCDIR) $(SRCDIR)ldeether.c $(DLPIFILES) $(LDEETHERLDFLAGS) -o $(OSARCHDIR)ldeether
```

Instead of linking all of Maiko, it just links in the `src/dlpi.c`
object file.


nbriggs (approved):
LGTM.

PR #268, opened 2021-01-21
build: Don't run ls -l when building ldeether.

waywardmonkeys:
Perhaps this was there as a reminder about it needing to be
`setuid`, but it shouldn't be in the default build output.


nbriggs (approved):
LGTM.  Didn't seem to add any value.

PR #269, opened 2021-01-22
More SUNDISPLAY / NOPIXRECT fixes.

waywardmonkeys:
This fixes most of the remaining issues, apart from llcolor.

### DIFF
--- a/src/initdsp.c
+++ b/src/initdsp.c
@@ -18,35 +18,20 @@
 #include <stdio.h>
 #include <unistd.h>
 
+#ifdef SUNDISPLAY
 #ifndef NOPIXRECT
-#ifndef DOS
-#ifdef XWINDOW
-#include <sys/time.h>
-#include <pixrect/pixrect_hs.h>
-#include <sunwindow/notify.h>
-#include <sunwindow/rect.h>
-#include <sunwindow/rectlist.h>
-#include <sunwindow/pixwin.h>
-#include <sunwindow/pw_util.h>
-#include <sunwindow/win_struct.h>
-#include <sunwindow/win_environ.h>
-#include <sunwindow/cms.h>
-#include <sunwindow/win_input.h>
-#else
 #include <sunwindow/window_hs.h>
 #include <sunwindow/cms.h>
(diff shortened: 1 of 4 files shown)


src/initdsp.c
```diff
@@ -18,8 +18,8 @@
 #include <stdio.h>
 #include <unistd.h>
 
+#ifdef SUNDISPLAY
 #ifndef NOPIXRECT
-#ifndef DOS
 #ifdef XWINDOW
```
nbriggs, line 23: This doesn't seem right, aren't  SUNDISPLAY and XWINDOW mutually exclusive?

waywardmonkeys: I thought so too! But ... I wasn't sure. There's no way to compile any of this, so I can take a best guess, but not sure what else.

waywardmonkeys:
As an aside, these changes are important if we want to support a new display type ... and a good example of that would be a hypothetical port running in the browser via WebAssembly.

src/initdsp.c
waywardmonkeys: Okay, I've removed the whole chunk of includes there. Spot inspection makes it seem like it would be okay. (But of course, we'll never compile that code again anyway...)

nbriggs:
Yes, the current setup is a mix of direct mapped frame buffer vs window system, where the window system is either SunWindows or X11, and then whether the Pixrect library was being used as an aid for SunWindows (or possibly, not sure on this...) or the frame buffer operations.

Underneath it all, as far as Lisp is concerned, is a memory mapped display and a cursor.  Although you can examine the bitmap that is the display, you're not supposed to see the cursor there.  Lacking (hardware/system software) support for that on DOS, at least, and possibly for some of the Sun hardware without cursor support,  is why there's a lot of really ugly stuff about taking the mouse (cursor) down.

waywardmonkeys:
I've just done a force push that fixes a couple of additional issues in `src/initkbd.c`.

nbriggs (approved):
Compiled on Solaris SPARC.  Otherwise, visual inspection seems reasonable.

PR #270, opened 2021-01-22
Fix compilation if !SUNDISPLAY && !XWINDOW && !DOS.

waywardmonkeys:
Some extra braces make this syntax more resilient. Go ahead and
fix a bit of formatting at the same time.


nbriggs (commented):
I am not sure how to evaluate this.

nbriggs (approved):
Formatting seems ok.  Did you try running it through clang-format (there's a .clang-format set of settings in the source tree) ?

waywardmonkeys:
The issue was just that without braces, the code came out to `if (...) else ...` so it wasn't valid.

PR #271, opened 2021-01-22
Begin to modernize optimized dispatch.

waywardmonkeys:
This feature was controlled by the compilation flag `OPDISP`
which would enable some bits of assembler on the x86 (ISC or DOS)
or some other specialized code on SPARC. On SPARC hardware, there
was a special compilation process that would preprocess the code
and generate dispatch tables.

We do this now when this feature is enabled using gcc's computed
gotos feature. This is available in clang and some other compilers.
Notably, it isn't present in Visual Studio.

This doesn't decrease our portability at all as this feature is
optional and it replaces specialized assembler code with C using
compiler extensions (making it cross-platform).

In doing this, we've removed a bunch of related code, however,
it is likely that other pieces yet remain and will be removed
in subsequent commits as we clean things up and refine them.

This feature remains disabled by default for now.


nbriggs (approved):
LGTM.   The computed gotos are available in the Oracle/Sun Developer Studio compilers as well as clang and gcc.   Gcc hints that one should consider -fno-gcse when using computed gotos with optimization turned on, but I haven't looked at whether that impacts this code.

PR #272, opened 2021-01-23
Remove profile.h

waywardmonkeys:
This no longer had anything to do with profiling and was
only doing some defines for the switch case block addresses,
which we no longer need since there's no longer optional
asm generated for them.


bin/legacy/makefile-sunos3.mc68020-multi
```diff
@@ -32,7 +32,7 @@ $(OBJECTDIR)xc.o :  $(SRCDIR)xc.c $(INCDIR)lispemul.h  $(INCDIR)emulglobal.h  \
 	 $(INCDIR)tosfuncallmacro.h  $(INCDIR)inlinedefsC.h   \
```
nbriggs, line 32: most of the `*inline*.h` and `*fast*.h` files don't exist anymore either.

nbriggs (approved):
LGTM.

PR #273, opened 2021-01-23
Remove special rules for building xc.c.

waywardmonkeys:
This also removes the special `DISP*` flags that were used for
building some files, including `xc.c`.


nbriggs (approved):
LGTM.  The listed dependencies for `xc.o` (and others) should get an update at some point.  It looks as though we're also pretty close to being able to have a single Makefile.

waywardmonkeys:
I have local changes that get us most of the way to auto-generated dependencies. But they have various issues that need to get sorted out first ... so we'll get there.

PR #274, opened 2021-01-24
Remove old setup_table optimizer workarounds.

waywardmonkeys:
These were for old compilers aren't needed any longer with
the new `OPDISP` code arrangement and modern compilers.

### DIFF
--- a/src/xc.c
+++ b/src/xc.c
@@ -1046,15 +1046,6 @@ nextopcode:
     case 0377:
     case377:
       CLARITHEQUAL;
-#ifdef OPDISP
-#ifdef ISC
-    case 0400:
-      goto setup_table; /* to defeat optimizer, so optable exists */
-#elif (DOS && OPDISP)
-    case 0400: goto setup_table;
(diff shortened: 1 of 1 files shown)


src/xc.c
```diff
@@ -1046,15 +1046,6 @@ op_ufn : {
...
-    case 0400:
-      goto setup_table; /* to defeat optimizer, so optable exists */
-#elif (DOS && OPDISP)
-    case 0400: goto setup_table;
-#endif /* ISC */
-
-#endif /* OPDISP */
 
 #ifdef I386
     /* to defeat the damn optimizer, make it look like */
```
nbriggs, line 1051: Likewise this section below, which is trying to make references to the error labels...

waywardmonkeys: That section might actually be needed if the corresponding code came back into play! But it won't ... but I planned to remove it when I remove the other related code.

nbriggs (approved):
LGTM

PR #276, opened 2021-01-24
Remove assembly fn call optimizations.

waywardmonkeys:
These were for old platforms and not used in recent history.

There's no NOASMFNCALL flag as there's no more asm fn call code.


nbriggs (approved):
LGTM

PR #277, opened 2021-01-24
Remove OPCODEFAIL / FIXTOS1

waywardmonkeys:
These had been used on Sun3 in the ancient past, but not after
that in the current code.


nbriggs (approved):
Looks good.

PR #278, opened 2021-01-24
Remove -fwrapv compiler flag.

waywardmonkeys:
This was making sure that the overflow in some our opcode
implementations was visible to the C code so that overflow
detection would work in gcc and clang.

This is now handled on those compilers by using some built-ins
that detect overflow.

The `-fwrapv` should no longer be needed.

Closes Interlisp/medley#90.


nbriggs (approved):
LGTM.  Tested on both macOS 32-bit clang and SPARC 64-bit gcc.

PR #279, opened 2021-01-24
Make sure OLD_CURSOR code is only for SUNDISPLAY.

waywardmonkeys:
This is work in progress on Interlisp/medley#84.


nbriggs (approved):
LGTM.

PR #280, opened 2021-01-24
Remove unnecessary (and incorrect) include in unixcomm.c

nbriggs:
unixcomm.c conditionally included `"inlinePS2.h"` -- which hasn't been named that in a long while, and is not necessary anyway.


waywardmonkeys (approved):
LGTM

PR #281, opened 2021-01-24
Clean up extraneous semi-colons, break statements, and braces

nbriggs:
Changed the POP_SUBR_ARGS definition to do { } while(0) style to avoid warnings for extra semicolons.
Removed extraneous braces in a number of case statements.
Removed extraneous (unreachable) break statements.
Added a few blank lines to improve formatting in switch cases.


waywardmonkeys (approved):
LGTM

PR #282, opened 2021-01-25
Fix NOPIXRECT in llcolor.

waywardmonkeys:
This code is all currently set up to be used if `COLOR` is
defined.

If we were to start supporting color under X11, this would
have to change, but so would much of the rest of the file.


nbriggs (approved):
LGTM

PR #284, opened 2021-01-25
Remove OLD_CURSOR from our builds.

waywardmonkeys:
This is no longer used under X11, which all of our builds are using.

Closes Interlisp/medley#84.


nbriggs (approved):
LGTM

PR #285, opened 2021-01-25
Remove extra DOS build files.

waywardmonkeys:
None of these work and we currently have no path towards
compiling for DOS that would use any of them.


nbriggs (approved):
Seems reasonable.  Nothing to test here.

PR #287, opened 2021-01-25
Don't enable NOPIXRECT when XWINDOW is defined.

waywardmonkeys:
This used to be required, but has been fixed now.


nbriggs (approved):
LGTM.

PR #290, opened 2021-01-25
osmsg now uses MAIKO_HANDLE_CONSOLE_MESSAGES

waywardmonkeys:
This was checking that neither `DOS` nor `XWINDOW` were defined,
but this needs a more specific flag for this feature so that it
doesn't get enabled when `!DOS && !XWINDOW`.

### DIFF
--- a/src/osmsg.c
+++ b/src/osmsg.c
@@ -19,23 +19,21 @@
 
 #include "version.h"
 
-#ifndef DOS
+#include <setjmp.h> /* Needed for jmp_buf for timeout.h */
+
+#include <errno.h>
 #include <pwd.h>
+#include <stdio.h>
 #include <sys/file.h>
+#include <sys/ioctl.h>
 #include <sys/select.h>
 #include <sys/stat.h>
-#include <sys/types.h>
-#include <unistd.h>
-#ifndef XWINDOW
-#include <sys/ioctl.h>
-#endif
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM.  The only time that it makes sense to have Lisp take the console log messages is when it is running on the raw framebuffer device that would otherwise be the system console, and it needs to capture logging messages that would otherwise print there, so that they don't disrupt the screen image and can be retrieved and played back into a Lisp window.

waywardmonkeys:
Yep, but if someone were hypothetically writing a different display backend, then it gets annoying when things don't compile without one of `XWINDOW` or `DOS` or `SUNDISPLAY`. :) And I didn't want to tie that functionality to `SUNDISPLAY` ...

waywardmonkeys:
(And that code should be updated to use `termio` anyway...)

nbriggs:
Yeah, I wasn't suggesting that it should be any different that what you did, just throwing in a bit of the rationale for why it existed at all.

masinter:
what does URAID use except termio?

waywardmonkeys:
@masinter The code in `osmsg.c` is still using pre-`termio` `ioctl` calls from the `sgtty` days. That code isn't used, so it isn't a priority to update. This is unrelated to URAID. URAID uses `stdio.h`, no `termio` calls.

PR #293, opened 2021-01-25
Fix bbtsub.c compilation without XWINDOW or DOS.

waywardmonkeys:
Without those, it wasn't including `devif.h`, which is where
we define `min` and `max`. I make an assumption here that any
new display would be using the current display device support
code.

### DIFF
--- a/src/bbtsub.c
+++ b/src/bbtsub.c
@@ -76,10 +76,10 @@ extern IOPAGE *IOPage68K;
 #include "dbprint.h"
 
 
-#if (defined(DOS) || defined(XWINDOW))
+#if !defined(SUNDISPLAY)
 #include "devif.h"
 extern DspInterface currentdsp;
-#endif /* DOS || WXINDOW */
+#endif /* SUNDISPLAY */
 
 #ifdef COLOR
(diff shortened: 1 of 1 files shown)


nbriggs:
We have quite a selection of min/max definitions
```
inc/cdrom.h:#define min(a, b) ((a <= b)?a:b)
inc/devif.h:#define min( a, b ) (((a)<(b))?(a):(b))
inc/locfile.h:#define min(a, b) ((a <= b)?a:b)
inc/mnxdefs.h:#define min( a, b ) (((a)<(b))?(a):(b))
src/gcarray.c:        #define min(a,b)		((a > b)?b:a)
src/gccode.c:#define min(a, b) ((a > b) ? b : a)
src/gcfinal.c:#define min(a, b) ((a > b) ? b : a)
src/picture.c:#define min(x, y) (((x) > (y)) ? (y) : (x))
```

waywardmonkeys:
Also `IMIN` :) ... but it is in my `NOTES` file as something to address in the future.

nbriggs (approved):
LGTM

PR #294, opened 2021-01-25
Remove bbtsub.h, duplicate of bbtsubdefs.h.

waywardmonkeys:
`bbtsubdefs.h` and `bbtsub.h` were close to the same, but
`bbtsub.h` was more complete and included by more things.


CMakeLists.txt
```diff
@@ -236,7 +236,6 @@ SET(MAIKO_HDRS
     inc/arraydefs.h
     inc/array.h
     inc/bb.h
-    inc/bbtsubdefs.h
     inc/bbtsub.h
```
nbriggs, line 239: How about removing bbtsub.h and leaving bbtsubdefs.h (to match the other xxxdefs.h) which also has the #ifdef guard in it ?

waywardmonkeys: I'll update this to do that.

waywardmonkeys: Updated.

PR #295, opened 2021-01-25
Move optable into static, read-only storage.

waywardmonkeys:
Previously, we had a global `optable`, an alias to it, `table`,
and code to initialize it within `dispatch()`. This code would
initialize the whole thing to either `op_ufn` or `native_check`
and then fill in the entries for the individual opcodes.

Now, we have a function-static `optable` which is statically
initialized and placed (by the compiler) into read-only storage.
This eliminates having generated code within `dispatch()` that
initializes it.

Additionally, it was previously 512 elements long with the second
half of it being all `native_check` entries. We index it with a
byte though, so we only need 256 entries, and we can lose the
`native_check` stuff.

### DIFF
--- a/src/xc.c
+++ b/src/xc.c
@@ -163,10 +163,6 @@ extern int Event_Req; /* != 0 when it's time to check X events
                                                  (e.g. Suns running OpenWindows) */
 #endif                /* XWINDOW */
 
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM.

PR #296, opened 2021-01-25
Build .o files for executables separately.

waywardmonkeys:
We need to build `.o` files for the executables separately from
their link step. This lets us have the `CFLAGS` only get used
by the compilation step and not the link step (which uses
`LDFLAGS` or a variant).

This is a prerequisite for doing automatic dependency tracking.


nbriggs (approved):
LGTM.  By the way, the (Oracle) Developer Studio compiler also generates dependency information, though using `-xMMD` rather than `-MMD` (and -xMF rather than -MF for other-than-default output file).

PR #297, opened 2021-01-25
Remove unix password checking from uutils.c

nbriggs:
unistd.h is the normal source for the prototype form crypt(), however the Solaris headers require` _XOPEN_SOURCE 700` or `__EXTENSIONS__` before they will expose it, and compiling with `--std=gnu99` doesn't give us the `__EXTENSIONS__`.


waywardmonkeys:
While this looks good to me for the specific thing you're updating, but:

* `getpwnam` doesn't return the crypted password when using a shadow file.
* `crypt` should just get passed the whole crypted password as the salt since it also usually includes algorithm information these days.
* We could use `getspnam` to get the shadow password entry, but then you'd need to be running setuid.
* We could use PAM, but again to verify against a shadow password entry,  you'd need to be running setuid.
* We could remove this functionality.

For the 2nd item above, we'd want to remove the salt variable and change one line to:

```
if (strcmp(crypt(pass, pwd->pw_passwd), pwd->pw_passwd) == 0)
```

nbriggs:
Yeah.  It's tied to a SUBR that does get called so  I suppose we could take the DOS path and always return T.

waywardmonkeys:
We should make notes of things that should be removed or modified on the lisp side.

nbriggs:
I suspect that this code hasn't worked in a LONG time.  Probably starting a discussion item would be the easiest way to track this.

waywardmonkeys:
Sure.

waywardmonkeys (approved):
LGTM. I still think we need a way to communicate these changes back to medley for code to get updated there, but that's probably separate from this.

PR #298, opened 2021-01-26
Sprinkle some `const` on `char *` params.

waywardmonkeys:
This is enough to let us move `il_string` in `loopsops.c`
into read-only storage.


nbriggs (approved):
LGTM.

PR #299, opened 2021-01-26
Use FD_CLR when disabling keyboard for SUNDISPLAY.

waywardmonkeys:
In the old code (prior to b234064d), this was:

```
   LispReadFds &= ~(1 << LispWindowFd);
```

This was inadvertently converted to `FD_SET`, but should have
been `FD_CLR`.


nbriggs (approved):
oops.  LGTM now.

PR #303, opened 2021-01-26
make makefile-tail cleanup target remove everything

nbriggs:
The makefile-tail cleanup target had accumulated a few missing files that it should have cleaned up, however rather than listing each individual file to be removed we can remove the object and executable directories.  The directories are recreated as necessary.


waywardmonkeys (approved):
LGTM

PR #304, opened 2021-01-26
Remove provision for offline assembler peephole optimizations on SunOS3/4

nbriggs:
We no longer need to run a separate peephole optimization pass over the
compiler generated or hand-coded assembler for SunOS3 and SunOS4 on 68K or SPARC

We remove the preprocessor symbols and code controlled by them
	SUN4_OS4_IL
	SUN3_OS3_IL
	SUN3_OS4_IL
and	SUN3_OS3_OR_OS4_IL


waywardmonkeys (approved):
LGTM

PR #305, opened 2021-01-27
cmake: Don't link maiko into setsout, tstsout.

waywardmonkeys:
These only need byteswap utils, not the fully maiko set of
sources. This makes this behave like the old make-based
build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,17 +43,13 @@ IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64" OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL "x86" OR
    CMAKE_SYSTEM_PROCESSOR STREQUAL "i386" OR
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM.  I've checked all three compilers we use (clang, gcc, and Developer Studio) and all of them define` __BYTE_ORDER__` to be one of `__ORDER_BIG_ENDIAN__` or `__ORDER_LITTLE_ENDIAN__` -- so I would expect that we'd be able to do away with defining `BYTESWAP` on the command line and derive it easily in `maiko/platform.h` (and make the byteswap object mostly empty when it's not providing any services).

waywardmonkeys:
I'd love to see that particular cleanup done! If you want to take it on ... I'd be good with that too. :)

nbriggs:
I'll look at it, at least putting in the automatic detection.   In the best of all possible worlds the source wouldn't be all `#ifdef BYTESWAP...#else...`, and would use appropriate access functions (along the lines of the BSD `byteorder(9)` functions), but small steps...

waywardmonkeys:
Small steps are good! Look how far things have come in the last 2 months in small steps. :)

PR #306, opened 2021-01-27
Make the table for reversing bits const

nbriggs:
... and remove a reference to the reversed bits table from a place that it was not used.


waywardmonkeys (changes requested):
There's another line in `src/mnxmeth.c` that needs updating (two extern decls).

nbriggs:
Dang.  Fixed that.  Moved it to a single reference early on in the file.

waywardmonkeys (approved):
LGTM

PR #307, opened 2021-01-27
Make (another) mask_array const.

nbriggs:


waywardmonkeys (approved):
LGTM. But formatting would be nicer as:

```
static const int mask_array =  {
    1, ....
    0x1ff, ...
};
```

nbriggs:
Better?  The old formatting was just what clang-format (version 4.0.0) produced, using style "Google" (but with line length 100).  I haven't spent any time tweaking the parameters but if you have suggestions I'd be willing to consider them.

waywardmonkeys (approved):
LGTM

PR #308, opened 2021-01-27
cmake: Improve build structure for multiple display implementations.

waywardmonkeys:
We no longer build a separate `maiko` library that gets linked
into `ldex`. It was the only thing linking it, so now we just
build those files as part of `ldex`.

We only build `ldex` when X11 is requested (which it is by default).

If we add a new display implementation, this structure will support
that by adding a new `ldewhatever` and it will build a new binary
that the `lde` launcher can be updated to support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,23 +18,50 @@ ENDIF()
 
 INCLUDE(CheckLibraryExists)
 CHECK_LIBRARY_EXISTS(m fmod "" NEED_LIB_M)
-
-FIND_PACKAGE(X11 REQUIRED)
+IF(NEED_LIB_M)
+  SET(MAIKO_LIBRARIES m)
+ENDIF()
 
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
This change LGTM, but testing on Solaris/SPARC showed up a problem from b6f46cd1f05d7601e7ec8eaa08f7491a32e35803 --
```
[ 95%] Linking C executable ldex
Undefined                       first referenced
 symbol                             in file
swapx                               CMakeFiles/ldex.dir/src/fvar.c.o
reversedbits                        CMakeFiles/ldex.dir/src/xcursor.c.o
ld: fatal: symbol referencing errors
```
because we need the `byteswap.o` even on big-endian machines.

waywardmonkeys:
Okay, I'll submit a new PR for the byteswap thing. :)

PR #309, opened 2021-01-28
clang-format: AllowShortBlocksOnASingleLine only if empty

nbriggs:
This will cause clang-format to produce
```
while (true) {}
while (true) {
  continue;
}
```
rather than forcing the empty block to two lines, or allowing the `continue;` and brace to merge up.


waywardmonkeys (approved):
LGTM

PR #310, opened 2021-01-28
cmake: Always build byteswap.c.

waywardmonkeys:
This is used on all platforms, not just ones where we need
byteswapping.


nbriggs (approved):
LGTM.

PR #313, opened 2021-01-28
make: Fix LDFLAGS ordering to fix cygwin build.

waywardmonkeys:
Cygwin has a linker that requires the libs to come last.

Closes Interlisp/medley#170.


waywardmonkeys:
This is pending @masinter saying that this fix works.

nbriggs (approved):
@masinter reports that this fixes Interlisp/medley#170

PR #314, opened 2021-01-28
Improve tty.c, fix compilation on Cygwin.

waywardmonkeys:
`EXTA` and `PENDIN` are not POSIX, so let's not use them.

At the same time, correct the baud symbol translation to match
the constants in medley's `CHARDEVICE`.

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -51,7 +51,7 @@ void tty_open(void)
       options.c_iflag &= ~(IMAXBEL|IXOFF|INPCK|BRKINT|PARMRK|ISTRIP|INLCR|IGNCR|ICRNL|IXON|IGNPAR);
       options.c_iflag |= IGNBRK;
       options.c_oflag &= ~OPOST;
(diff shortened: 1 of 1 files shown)


nbriggs (approved):
LGTM by visual inspection (vs CHARDEVICE)

waywardmonkeys:
So, `tty.c` is meant to handle the needs of `DLTTY` which was the implementation for the Dandelion hardware and not supported in the SunOS version. But someone was working on this code to add that support... but the baud constants used in the last version of this file are from `DLTTY` Which differ from `CHARDEVICE`. But `DLTTY` was deleted...

a historical copy can be found at https://github.com/Interlisp/history/blob/master/202011-lisp/library/DLTTY

waywardmonkeys:
I've updated this to follow the `DLTTY` constants until we decide what to do about `DLTTY`, `CHARDEVICE`, and the rs232 code.

PR #315, opened 2021-01-28
Remove various unused externs.

waywardmonkeys:
* `cursor_mask_gc` is removed as it isn't used anywhere.
* `IconPixmap` is made into a static in `xmkicon.c` rather than
  being defined in `xlspwin.c`.


nbriggs (approved):
LGTM.  Re: Bitmap_Pad -- not explicitly initialized, but 0 because it's a global.

PR #316, opened 2021-01-28
mnwevent: Fix copy/paste error in log message.

waywardmonkeys:
This gets it out of search results when looking for getXsignaldata.


nbriggs (approved):
LGTM

PR #319, opened 2021-01-29
Remove cursor_source_gc.

waywardmonkeys:
This was only used in commented out code, which is now removed
as well. In that code, it was used to set the function to `GXcopy`,
but it would already have been that as that's the only thing
we set function to.

This also lets us remove the `gcv` code in `xcursor.c`, since
we no longer need those values there.

### DIFF
--- a/src/xcursor.c
+++ b/src/xcursor.c
@@ -28,9 +28,6 @@
 
 extern IOPAGE *IOPage;
 
-extern XGCValues gcv;
-
-GC cursor_source_gc;
 XColor cursor_fore_xcsd, cursor_back_xcsd, xced;
 extern Colormap Colors;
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
After this, there's only one use of the `AIX` define, and that's only for systems where we capture console output, which isn't any of our current systems.

So, I'll finally be able to remove the `-DAIX` from all build systems that has been annoying me for ages.

nbriggs (approved):
OK.  This was almost all related, I think, to inverting the cursor bitmap colors when you invert the screen color -- except that I verified that that was NOT what the D-machines did, so it was commented out, and I fixed the screen invert functionality so that it worked properly.

PR #320, opened 2021-01-29
Remove AIX define from Linux, Cygwin, Solaris, OpenBSD.

waywardmonkeys:


nbriggs (approved):
At least Solaris, FreeBSD, and macOS have the TIOCCONS ioctl to "become virtual console".  I haven't checked on Linux.

PR #321, opened 2021-01-30
Add an opcodes enum with names for each opcode.

waywardmonkeys:
This is based on the info in `medley/sources/LLCODE`.


inc/opcodes.h
```diff
@@ -20,28 +20,268 @@
...
+    opc_WRTPTR0TAG = 55,
+    opc_MISC7 = 56,
+    opc_DOVEMISC = 57,
+    opc_EQL = 58,
+    opc_DRAWLINE = 59,
+    opc_STORE_N = 60,
+    opc_COPY_N = 61,
+    opc_RAID = 62,
+    opc_SLRETURN = 63,
+    opc_IVAR1 = 64,
```
nbriggs, line 88: I think the convention for the variable slots is that they're numbered from 0.  The reference being looking at the results of (INSPECTCODE 'fn-name)

waywardmonkeys: Updated for that.

nbriggs (approved):
LGTM

PR #324, opened 2021-01-30
Fix missing initialization in int_panic_init()

nbriggs:
panic_action.sa_mask and .sa_flags initialized twice, corresponding fields
in ignore_action not initialized.


waywardmonkeys (approved):
LGTM. Copy and paste error on my part.

PR #325, opened 2021-01-30
Remove redundant calls to XFlush

nbriggs:
Calls to flush_display_region() are responsible for calling XFlush()
and it is pointless to follow them up with another call.


waywardmonkeys (approved):
LGTM. These call `flush_display_region`, which goes through the display interface calling `bitblt_to_screen` which, on X, is `clipping_Xbitblt`, which does the flush when needed.

PR #326, opened 2021-01-30
Define RM in makefile if it is not already defined

nbriggs:
Some versions of "make" do not predefine make variable RM (looking at you, FreeBSD)
so if it has NOT been defined we define it as "rm -f"

Also, update the cleanup target to use $(RM) rather than refering directly to "rm".


waywardmonkeys (approved):
LGTM. Sad news about FreeBSD. I had checked Solaris make as I figured it would be the most backwards.

PR #327, opened 2021-01-31
Fix off-by-one error in clipping_Xbitblt

nbriggs:
The LowerRightX and LowerRightY were calculated as the X and Y coordinates of the
pixel to the right and below the last visible pixel, but are compared with the
last actually visible pixel in the source. They need an extra -1 offset to correct.

Add comments to explain what's going on.  Fix the signature of the function to reflect
that the "dummy" passed for the display region is a pointer not an integer.


waywardmonkeys (approved):
I think this is right.

PR #328, opened 2021-01-31
Misc7 opcode flush correct region

nbriggs:
The MISC7 opcode was not accounting for the offset from the base when
calculating which (16-bit) word should be flushed to the actual display.


waywardmonkeys (approved):
LGTM. So this function is doing a bit manipulation on part of the bitmap, and it may invert, erase, or set. It is doing that at `base + offset`, but we were inadvertently only doing the flush at the base, not the region that had changed. (Just confirming my understanding.)

nbriggs:
Your understanding is correct.  It can modify exactly 1 bit in the bitmap, but was flushing the wrong word to the display.

PR #329, opened 2021-01-31
Remove support for the Sun386i.

waywardmonkeys:
This was a machine running SunOS 4, which we no longer support.

It had an odd display controller and a compiler with issues that
required workarounds. Those compiler issues aren't an issue in
today's world, so we don't need to keep the workarounds present
for reference.

It had a bit of inline assembly, but that is still present in
other files for other platforms.


nbriggs (approved):
LGTM.  This symbol controlled an odd mix of things that were related to the display controller and weirdnesses in the optimization of the gcc that Sun distributed with for this platform (e.g., definition for Addr68k_from_LADDR, etc.)

I'm pretty sure that in a future PR the GCC386 stuff can go away too.

waywardmonkeys:
GCC386 will get removed this coming week, I think!

PR #330, opened 2021-01-31
Detect byte order from preprocessor definition

nbriggs:
Rather than having to specify or omit `-DBYTESWAP` in the Makefile
fragment for each system we can detect whether the system requires
byte swapping  based on the `__BYTE_ORDER__` C preprocessor definition.

This does not replace use of the existing `BYTESWAP` define in the sources by,
for example, `MAIKO_BYTESWAP`.


waywardmonkeys (changes requested):
In the `CMakeLists.txt`, this block can be removed:

```
IF(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "x86" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "i386" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "i686" OR
   CMAKE_SYSTEM_PROCESSOR MATCHES "arm.*" OR
   CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
  LIST(APPEND MAIKO_DEFINITIONS
    "-DBYTESWAP"
  )
ENDIF()
```

waywardmonkeys (approved):
LGTM

PR #331, opened 2021-02-01
Clean up MAIKO_HANDLE_CONSOLE_MESSAGES some.

waywardmonkeys:
* `LOGINT` is only useful when `MAIKO_HANDLE_CONSOLE_MESSAGES`
  is set.
* Only define some variables if `MAIKO_HANDLE_CONSOLE_MESSAGES`
  is set. Flag them as static.
* Remove setting `LOGINT` from the build system as it isn't
  needed because we aren't handling console messages.


nbriggs (approved):
OK.  I think that to compile on macOS one would need to include `<sgtty.h>`, and one can get most of the way to compiling on Solaris with `<fcntl.h>`, and `<termios.h>`, but I don't think there's a functional definition of the TIOCCONS ioctl.  I suspect that removing all this code is probably more useful than trying to resurrect it.

PR #332, opened 2021-02-01
Fix compilation with MAIKO_HANDLE_CONSOLE_MESSAGES.

waywardmonkeys:
* Switch from sgtty over to termios. All of the various
  separate values that were copied individually before
  can now be done at once.
* Fix compilation error with missing `&`.
* Fix compilation error with missing return value.

I haven't tested that this works as I don't know how to do that,
but at least it compiles now.

### DIFF
--- a/src/osmsg.c
+++ b/src/osmsg.c
@@ -31,6 +31,7 @@
 #include <sys/statvfs.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <termios.h>
 #include <unistd.h>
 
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
This should, in theory, work on macOS and Linux if enabled at build time. Solaris doesn't have `TIOCCONS`, so it won't work there.

nbriggs (approved):
LGTM.  I have to figure out what needs to happen on the Lisp side, but I think in theory, on a Mac, it should receive the "panic" level (0) messages sent by, for example, `syslog -s -l 0 This is an emergency level message to all connected terminal sessions.`

PR #333, opened 2021-02-03
Fix the unixfork/unixcomm code to deal with pid over 65535

nbriggs:
Process ids can now be over 65535 so expand from 2-byte to 4-byte
in the packet exchange.  Rearrange some of the code to make it a little
more comprehensible.  Update (somewhat) the documentation.
The packet format should be changed/reordered, and the code should be verified
against the Medley Lisp code that drives it.  Some of the commands don't appear
to be used, or may be used inappropriately.

### DIFF
--- a/inc/unixfork.h
+++ b/inc/unixfork.h
@@ -1,5 +1,4 @@
 #ifndef UNIXFORK_H
 #define UNIXFORK_H 1
 int fork_Unix(void);
-int ForkUnixShell(int slot, char *PtySlave, char *termtype, char *shellarg);
 #endif /* UNIXFORK_H */
--- a/src/unixcomm.c
+++ b/src/unixcomm.c
@@ -56,14 +56,13 @@ Unix Interface Communications
 #include "byteswapdefs.h"
 #include "commondefs.h"
 
-static __inline__ int SAFEREAD(int f, unsigned char *b, int c) {
-  int res;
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
I am not clear on why we have this forking process at all. What is it for?

I'm short on time this week due to some other commitments.

It would be easier if some of these changes were in either separate commits within the same PR or in separate PRs so that the changes related to the PID are clearly visible vs things like cleaning up the read function.

nbriggs:
It's trying to never block the main interpreter loop by executing things that could take unlimited time.  It does it by turning it into async I/O through a pipe to (another fork of) a small already forked process.   Additionally to the blocking issue, you didn't want to be forking a large process in order to then `exec()` a small(er) one -- you'd frequently run out of memory (or swap space) when the large process was copied on the `fork()` call.  So, it arranges for a tiny process to be there then starts the rest of the system, and then forks the service processes from the small one on command.

waywardmonkeys (approved):
LGTM. I wish we could re-order the bytes so that the values were together, but until then, this looks fine. (Might also look at using a regular C99 inline rather than `__inline__`.)

nbriggs:
I'll change the `__inline__`, before I merge it.  We should certainly reorder the bytes -- I'll do it in another PR, though.

PR #334, opened 2021-02-04
Cleanup EVAL macro definition

nbriggs:
Reformat and make use of PTRMASK & swapx() as appropriate.
It's likely that nnewframe() could be reworked to avoid needing
to unconditionally swapx() the result and, like native_newframe()
return the appropriate value as the result.

### DIFF
--- a/inc/tosfns.h
+++ b/inc/tosfns.h
@@ -545,110 +545,89 @@ op_fn_common:								\
 /******			 EVAL				 ********/
 /****************************************************************/
 #ifndef BIGATOMS
-#define EVAL								\
-  {									\
-    LispPTR scratch;							\
-    register LispPTR work;						\
-    register LispPTR lookuped; /* keep looked up value */		\
-									\
-    switch(TOPOFSTACK & SEGMASK)					\
-      {									\
-	case S_POSITIVE:						\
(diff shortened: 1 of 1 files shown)


waywardmonkeys (approved):
Just visually looking and it seems fine to me.

There's a related change (for another PR) to be done to make `N_OP_fvarn` use `swapx` as well. I'd rather see something like `maiko_read_u16` and the like though that handle swapping when needed. But I guess this isn't endian-related swapping, since it happens all the time?

nbriggs:
Correct, this is not an endian-related swap.  It pokes two values through a `DLword *` to access the 0th and 1st elements of what is really a `LispPTR`, and then swaps the halves of what it just constructed (again) to make it the acceptable `LispPTR` value in the outside (in almost all the cases...)

nbriggs:
I wish that CPP comprehended (#if) directives within a #define macro definition.

waywardmonkeys:
And I wish that some of our `#define` blocks were inline functions. :)

nbriggs:
It's quite a bit of rewriting to convert the macros to inline functions, mostly because of the gotos for dispatch and the common UFNs.

PR #335, opened 2021-02-07
Address older versions of C compiler which don't have __BYTE_ORDER__

nbriggs:
Older versions of gcc, such as 4.0.1, found on a PowerPC macs running 10.5, don't have `__BYTE_ORDER__ ` but do define `__BIG_ENDIAN__` (or, presumably on other hardware, `__LITTLE_ENDIAN__`) so we can add checks for those if the preferred method for detecting byte order isn't available.

### DIFF
--- a/inc/maiko/platform.h
+++ b/inc/maiko/platform.h
@@ -137,6 +137,9 @@
 #  define MAIKO_ARCH_DETECTED 1
 #endif
 
+/* Modern GNU C, Clang, Sun Studio  provide __BYTE_ORDER__
+ * Older GNU C (ca. 4.0.1) provides __BIG_ENDIAN__/__LITTLE_ENDIAN__
+ */
 #if defined(__BYTE_ORDER__)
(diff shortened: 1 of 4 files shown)


nbriggs:
The first step in adding the makefile for PowerPC Mac OSX 10.5 was to detect the byte order correctly.
There's no CMake distributed for that system, so we'll rely on a makefile segment in the original style build.

inc/maiko/platform.h
```diff
@@ -145,6 +145,10 @@
 #  else
 #    error "Unknown byte order"
 #  endif
+#elif __BIG_ENDIAN__ == 1
```
waywardmonkeys, line 148: Could you add a comment here that this is for old gcc?

waywardmonkeys (approved):
LGTM with one nit.

inc/maiko/platform.h
nbriggs: Sure can.  I don't know in which version of GNU C or Clang they introduced `__BYTE_ORDER__` but I do know when it _didn't_ have it.

PR #336, opened 2021-02-09
Remove OEXT from makefile system.

waywardmonkeys:
This was originally here to make something in the Apollo toolchain
happy, but it hasn't been needed in ages, and was inconsistently
applied.


nbriggs (approved):
LGTM.

PR #337, opened 2021-02-09
Favor C99 inline over __inline__.

waywardmonkeys:


nbriggs (approved):
These seem fine.  I'm not sure why these were originally singled out to receive `__inline__` in the first place, but they're probably hotspots.

PR #338, opened 2021-02-09
Remove GetNumber.

waywardmonkeys:
This isn't used and if it were, it wouldn't compile. It calls
`ufn`, which doesn't get compiled in either (unless `C_ONLY`
is defined, which it isn't, and if it were, that wouldn't
compile either due to a missing `StkLim0`).

A subsequent PR will remove `ufn` and related code.


nbriggs:
I think this (and code in ufn.c) predates the `goto ufnlabel` style that is currently used.  The only thing that's interesting in there is the stack setup to before calling the Lisp UFN.   I wish we had proper documentation for the stack structure (e.g., "residual basic frames", the various markers in the entries returned by nnewframe(), and so on)

nbriggs (approved):
LGTM.

PR #339, opened 2021-02-10
Remove cdrom&cdaudio support code.

waywardmonkeys:
This is not used in Medley (the MISCN-TABLE-LIST doesn't include
the required constants), nor does anything use or reference this
code. It also won't compile as it needs some routines that are
missing from our code.


nbriggs (approved):
LGTM

PR #340, opened 2021-02-10
Remove Cldeetr.c.

waywardmonkeys:
This was like `ldeether`, but for SunOS4 with a color display. It
isn't needed any longer as on Solaris, we have `ldeether`, no color
display, and that code already works, unlike this one.


nbriggs (approved):
LGTM

PR #343, opened 2021-02-10
Experiment yield subr

masinter:
as discussed

### DIFF
--- a/inc/subrs.h
+++ b/inc/subrs.h
@@ -191,4 +191,6 @@
 #define user_subr_DUMMY	012
 #define user_subr_SAMPLE_USER_SUBR	00
 
+/* Experimental yield */
+#define sb_YIELD                        0322
 #endif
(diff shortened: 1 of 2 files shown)


nbriggs:
This is a mess and won't build.  It needs to be rebased on the current master, I think.

waywardmonkeys:
I'd like to hold off on this for a week (but keep it open).

The alternate display code that I'm working on will need something of this sort as well as it is using OpenGL (or Metal on macOS) (or WebGL in the browser), and it needs the ability to do a render periodically.

Also, running in the browser means that we can't have an infinite loop like we do now for `dispatch`.

masinter:
https://github.com/Interlisp/medley/pull/254#pullrequestreview-609197575

nbriggs:
What has #254 (unixprint) got to do with this?

nbriggs:
I think that the Link above (medley 254) is unrelated to this.

masinter:
it's been a month. I run with this always.

masinter:
FYI This PR was merged and the merge reverted, but it still shows up as Merged.

PR #344, opened 2021-02-10
Reformat (with clang-format 10.0) some header files

nbriggs:
...in preparation for editing macro definitions to wrap in do {} while (0)
or other adjustment to make them complete statements without unnecessary
semicolons.

This programmatic reformatting should not introduce any functional changes.


waywardmonkeys (approved):
My eyes will bleed reviewing this line by line, so a spot check and a skim says it LGTM.

nbriggs:
I hear you!  I'm separating it into a PR with no changes other than the programmatic reformatting so our eyes remain intact for the next change with the do..while (and re-reformat) -- that will be much easier to review with a "git diff -w ..." rather than reading through it on the web interface.

PR #345, opened 2021-02-11
Fix extra semicolon warnings in xc.c

nbriggs:
```
warning: empty expression statement has no effect; remove unnecessary ';' to silence this warning" [-Wextra-semi-stmt]
```
occurred 254 times in xc.c when warnings were turned up.

Review this commit with `git diff -w ...` (or check the "ignore whitespace" in the github web interface tools)
to make it easier to see changes other than indentation caused by
the addition of `do { ... } while (0)` around many macro definition bodies.

### DIFF
--- a/inc/arith.h
+++ b/inc/arith.h
@@ -30,19 +30,19 @@
 #define FLOATP_VALUE(dest) *((float *)Addr68k_from_LADDR(dest))
 
 #define N_GETNUMBER(sour, dest, label)                        \
-  {                                                           \
+  do {                                                        \
     dest = sour; /* access memory once */                     \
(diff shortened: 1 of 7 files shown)


waywardmonkeys (changes requested):
I'm not a fan of the changes with `do` without the brace. Some of the things have big chunks of code inside, so it isn't something that would be remembered when editing. (Like `ARITH_SWITCH`...)

nbriggs:
I was of two minds about those -- so sure, I'll add the extra braces.

waywardmonkeys (approved):
LGTM

PR #346, opened 2021-02-11
Remove -m32/-m64 args from makefiles.

waywardmonkeys:
We detect machinetype and build for the native CPU type as it
is, so we don't need to repeat this here in each makefile.

### DIFF
--- a/bin/makefile-cygwin.x86_64-x
+++ b/bin/makefile-cygwin.x86_64-x
@@ -1,7 +1,7 @@
 # Options for Linux, Intel x86_64 and X-Window
 
-#CC = gcc -m64 $(GCC_CFLAGS)
-CC = clang -m64 $(CLANG_CFLAGS)
+#CC = gcc $(GCC_CFLAGS)
+CC = clang $(CLANG_CFLAGS)
 
 XFILES = $(OBJECTDIR)xmkicon.o \
 	$(OBJECTDIR)xbbt.o \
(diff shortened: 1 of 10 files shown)


waywardmonkeys:
I'm adding `.mk` files that we can `include`, one for each compiler, so that I can configure things like dependency flags correctly for each compiler.

That would be easier if there weren't 32 / 64 bit flags being given, especially as I don't think they should be needed.

nbriggs (commented):
Generally, I agree with this.  I have a problem, though, in what ends up actually happening -- on my 10.11 (El Capitan) mac, the compiler default is 64-bit but it supports 32-bit as well and we have `bin/machinetype` reporting "386" rather than x86_64.   On the SPARC Solaris system, `bin/machinetype` reports "sparc" (not sparc64, or sparcv9), and 3 different C compilers with 7 different versions -- some of them default to 32-bit (sparcv8 instruction set), some default to 64-bit (sparcv9 instruction set).

I'd like to be able to more easily pick the compiler and options that get used.  It would also be great to have a consistent method to add extra (or change existing) options -- adding `-Wxxx` or `-fxxx`, change the `-On` and so on.

masinter:
on my mac mini I can do

arch -x86_64 ./makeright x

and compile maiko x86 binaries into darwin.x86_64/lde and ldex

nbriggs:
Yes, I can do `arch -x86_64 ./makeright x` on my MacBook Pro, but if I do `arch -i386 ./makeright x` it produces
>arch: posix_spawnp: ./makeright: Bad CPU type in executable

because all the executables used to create a 32-bit lde are in fact 64-bit executables, and for most other systems it's more like the `-m32/-m64` (or `--target=<target-triple>` for cross-compiling).

waywardmonkeys:
On macOS, it is usually `-arch i386` and `-arch x86_64` and `-arch arm64` ... I'm still thinking about what to do about this patch.

PR #348, opened 2021-02-11
Update more header include guards.

waywardmonkeys:
Some of these didn't have include guards, some  used old names
for them. Now they follow the current convention.


nbriggs (approved):
LGTM.  I think every file but the four `inln*.h` now have a current convention guard.

PR #349, opened 2021-02-12
Prevent duplicate printing of info lines in response to -info option

nbriggs:
The X options parser printed the info lines as well as the main options handler.
Add mention of the -info option in the -help usage information.

Fixes Interlisp/medley#41


waywardmonkeys (approved):
LGTM. I'd forgotten about this after having noticed it before.

PR #350, opened 2021-02-12
Allocate fptovp temporary table in heap

nbriggs:
Theres no reason to (stack) allocate a full-size (2 MB) temporary FPtoVP table when loading the sysout.
Heap allocate a sufficiently large (based on file size) table and then free it when the sysout is loaded.


waywardmonkeys (approved):
LGTM, but please don't squash the commits when merging. Should remain 2 separate commits.

waywardmonkeys:
Also, this fixes a bug that @masinter had with stack size on Cygwin (I think it was Cygwin).

nbriggs:
May also address Cygwin stack size from issue Interlisp/medley#103

PR #351, opened 2021-02-13
Fix missing includes in ldeether.c

nbriggs:
It was missing unistd.h (for various library/syscalls) and dlpidefs.h for the DLPI procedures.


waywardmonkeys (approved):
LGTM. I'll take your word for it, given that I lack any Solaris systems. :)

PR #354, opened 2021-02-13
Compiling with ethernet support requires etherdefs.h in certain places

nbriggs:
A number of files conditionally (if MAIKO_ENABLE_ETHERNET) refer to ethernet related functions that are declared in etherdefs.h, which was not (even conditionally) included.  This has been checked for compilation errors on Solaris with MAIKO_ENABLE_ETHERNET defined.


waywardmonkeys (approved):
LGTM

PR #355, opened 2021-02-14
Files should explicitly include the definitions they directly depend on

nbriggs:
... start by fixing the include files and the fallout in C source from
removing includes not directly needed by headers that previously included
them.


waywardmonkeys (approved):
Good idea. I have wanted to do similar. Did you use a tool?

nbriggs:
Re: "Did you use a tool" -- sort of.  "include-what-you-use", based on llvm/clang tooling.  From Google, I think.
It took a lot of iterations to get to a stable point.  It and and I disagree on some things, for example
```
medleyfp.h should add these lines:
#include <iso/math_c99.h>  // for isfinite

medleyfp.h should remove these lines:
- #include <math.h>  // lines 48-48

The full include-list for medleyfp.h:
#include <iso/math_c99.h>  // for isfinite
```
and you have to repeat with the selection of defined flags and use your judgement about leaving in things that aren't always going to be used.   It has a mode where it will also suggest forward declarations for structures so you don't have to include as many headers, but I turned that off for this.   I've tried it on the .c files, and the results are mostly not usable for our style of code, though they can give you a hint of where to look.

nbriggs:
One thing that doing this exposed was that there's a lot of room for improving how things are partitioned into the headers.

waywardmonkeys:
I haven’t made big changes to headers or suggested it as we need to fix dependency info in the old build system first...

nbriggs:
A manual update of the dependency info is not too hard to incorporate into the makefile-tail, however doing automated dependcies needs a decision on whether we're buying into GNUmake, or BSDmake (pmake), or sticking to POSIX make (2017 revision now specifies "include", so it's not completely useless).

PR #356, opened 2021-02-14
Remove unnecessary (re)definitions of TRUE/FALSE

nbriggs:
There are definitions for TRUE and FALSE in lispemul.h, which is included by most files that use them, so we can remove them from a number of (but not all) other files.


waywardmonkeys (approved):
YES

PR #358, opened 2021-02-16
Remove support for MS-DOS

nbriggs:
This is the first cut at removing support for MS-DOS as I don't think we're ever going to build there again

- remove (with "unifdef") all code depending on symbol DOS
- remove all .c and .h files only relevant to DOS
- remove these from CMake and traditional build
- remove the DOS makefile
- cleanup a few leftover defined(DOS) in logical expressions and fix endif comments
- remove a couple of assembler modules for DOS only


waywardmonkeys:
One thing that stopped me from doing this before was that I wasn’t sure if some of the code for handling drives in oaths might be useful for Windows ... thoughts?

nbriggs:
Yeah, you have a point there.  I wish we could dispense with drive letters, though.  Some of the ugliest code is in there.

nbriggs:
I'll take a shot at revising this to not strip out the file system related code.
I'm not sure what you envision on the input side if your new display code is implemented -- even if it's an X-like input there'll be a whole lot of specific code to deal with a Windows (rather than DOS) keyboard/mouse input.

waywardmonkeys:
For my display work, I'm using Sokol (https://github.com/floooh/sokol). Specifically, I am using `sokol_gfx.h` and parts of `sokol_app.h` (unless I can rework some other things and then I can use it directly).

This lets us run the same code on Unix, macOS, Windows, Web, and more.

I don't worry a lot about the existing code for handling keyboard and mouse input. :)

masinter:
My thoughts have involved a URL-based file system, and handling drive letters the way [whatwg/url](https://github.com/whatwg/url)  does.

nbriggs:
Cool.  Using the sokol_app "event-based keyboard, mouse and touch input" I guess.  It looks as though it will have issues of similar complexity to the way the X input events needed to be handled.

masinter (changes requested):
line 61 of (new) src/ether.c has `#endif /* DOS */` but the ifdef remover removed the wrong one: `#endif /* MAIKO_ENABLE_INTERNET */`

masinter (approved):
I think this might gate some other work

nbriggs:
@masinter -- I'll fix the endif problem.   I'm probably going to revise this substantially because Bruce asked that the file-system code handling drive letters be left for when/if he has a native Windows port.   It also has conflicting commits with the current master so I'll be needing to rebase and cleanup before merging.

PR #359, opened 2021-02-17
Clean up missing include file when compiling for INIT

nbriggs:
Missing prototypes in C99 mode are fatal in some compilation environments.  The files that conditionally reference `init_keyboard(...)` when compiling for INIT need to also conditionally include initkbddefs.h.

### DIFF
--- a/src/bbtsub.c
+++ b/src/bbtsub.c
@@ -76,6 +76,16 @@ extern IOPAGE *IOPage68K;
 #include "display.h"
 #include "dbprint.h"
 
+#ifdef INIT
+#include "initkbddefs.h"
+extern int  kbd_for_makeinit;
+
+#define init_kbd_startup   \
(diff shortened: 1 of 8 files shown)


nbriggs:
@masinter -- if you can try this on your mac mini and approve it if it works for you... I'll merge it.  Then I have another change to rename the makefile segments that are used for init and a fix to makeright so that it uses makefile-darwin.aarch64-init and so on so we can have the right INIT options for the right OS.

masinter (approved):

nbriggs:
Thanks.  I'm trying a loadup on my Mac (El Capitan, running ldeinit compiled 32-bit x86 mode), and I'm seeing an issue that looks like the keyboard event handling taking a segmentation fault.   I'll be looking into that more today, but it won't be until later this afternoon.

PR #360, opened 2021-02-18
Improve process for building ldeinit

nbriggs:


waywardmonkeys (approved):
This looks good to me. I think it means that nothing is building `setsout` and `tstsout` though, right?

nbriggs:
Correct -- they don't get built by default.  If you want them, then you run `./makeright x ../darwin.386/testsout` or whatever is appropriate for the architecture of the system you're running on.  `makeright` passes on the "extra" arguments directly to `make`.

PR #361, opened 2021-02-18
Add aarch64 Linux makefile

AbeJellinek:
`machinetype` reports `aarch64` on 64-bit ARM (at least under Docker for Mac), so we need a makefile for that configuration. It's a duplicate of `makefile-linux.armv7l-x` for now; that makefile isn't doing anything armv7-specific.

### DIFF
--- a/bin/makefile-linux.aarch64-x
+++ b/bin/makefile-linux.aarch64-x
@@ -1,0 +1,27 @@
+# Options for Linux, ARMv7 and X-Window
+
+CC = gcc $(GCC_CFLAGS)
+#CC = clang $(CLANG_CFLAGS)
+
+XFILES = $(OBJECTDIR)xmkicon.o \
+	$(OBJECTDIR)xbbt.o \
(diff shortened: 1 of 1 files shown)


nbriggs:
This will need an update after #360 is merged (or now, if you want) to change the default target as the file you based it on is about to change...
```
--- a/bin/makefile-linux.armv7l-x
+++ b/bin/makefile-linux.armv7l-x
@@ -24,4 +24,4 @@ LDELDFLAGS =  -L/usr/X11/lib -lX11 -lc -lm
 
 OBJECTDIR = ../$(RELEASENAME)/
 
-default        : ../$(OSARCHNAME)/lde 
+default        : ../$(OSARCHNAME)/lde ../$(OSARCHNAME)/ldex
```

waywardmonkeys (changes requested):
If you could make the change that @nbriggs suggested, I'd like to see this go in.

AbeJellinek:
Thanks for dealing with this!

PR #362, opened 2021-02-22
tidy: FindAvailablePty.

waywardmonkeys:
* Remove unused `Master` argument.
* Update documentation.
* Remove unused `MasterFD` variable in caller.
* Remove `slot` variable in caller and use `Master` in each case.
* Make `FindAvailablePty` a static function.


nbriggs (approved):
LGTM.  Reminds me I should go back and fix the layout of the command/response to put the PID bytes together.

PR #364, opened 2021-02-22
tidy: xrdopt.

waywardmonkeys:
* Remove unused `print_lispusage()`.
* Make some globals into function locals.
* Make the options table into a `static`. It can't be `const`
  because `XrmParseCommand` wants a mutable pointer.
* Remove unused `homeDB`.
* Stop using `caddr_t`, use `XPointer`. `caddr_t` is not POSIX
  and never made it past old obsolete BSD code.


nbriggs (approved):
LGTM.

PR #365, opened 2021-02-22
tidy: Remove last `caddr_t`.

waywardmonkeys:
This cast isn't needed here and `caddr_t` is obsolete / not POSIX.


nbriggs (approved):
So long caddr_t.  LGTM.

PR #366, opened 2021-02-22
XrmParseCommand and XrmMergeDatabases depend on destination database …

nbriggs:
…being initialized.

When they moved from global to local we missed that they needed to be explicitly initialized.


nbriggs:
Closes Interlisp/medley#212

PR #369, opened 2021-02-25
Fix printf format warnings in testtool.c

nbriggs:


nbriggs (commented):
@masinter -- the maiko (uraid) `doko` function currently prints the PC as octal -- given that you've changed the `INSPECTCODE` output to hex, should I change all uraid output to hex?

masinter:
image

pc 0x4
in hex almost everywhere

masinter (approved):

PR #370, opened 2021-03-03
Update README.md for simplified Mac instructions

masinter:
Now that you can get XQuartz for newer macs

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ $ cd maiko/bin
 $ ./makeright x
 ```
 
-* The build will (attempt to) detect the OS-type and cpu-type. It will build binaries `lde` and `ldex` in `../ostype.cputype` (with .o files in `..ostype.cputype-x`. For example, Linux on a 64-bit x86 will use `linux.x86_64`, while MacOS 11 on (new) Mac will use `darwin.aarch64`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,18 +33,17 @@ $ cd maiko/bin
 $ ./makeright x
 ```
 
-* The build will (attempt to) detect the OS-type and cpu-type. It will build binaries `lde` and `ldex` in `../ostype.cputype` (with .o files in `..ostype.cputype-x`. For example, Linux on a 64-bit x86 will use `linux.x86_64`, while MacOS 11 on (new) Mac will use `darwin.arm64`.
-* If you prefer using `gcc` over `clang`, you will need to edit the makefile fragment for your configuration (`makefile-ostype.cputype-x1) and comment out the line (with a #) that defines `CC` for `clang` and uncomment the line (delete the #) for the line that defines `CC` for `gcc`.
-* There is a cmake configuration (TBD)
+* The build will (attempt to) detect the OS-type and cpu-type. It will build binaries `lde` and `ldex` in `../ostype.cputype` (with .o files in `..ostype.cputype-x`. For example, Linux on a 64-bit x86 will use `linux.x86_64`, while MacOS 11 on (new) Mac will use `darwin.aarch64`.
```
waywardmonkeys, line 36: Since we're fixing this, say "on a (new M1) Mac" ...

waywardmonkeys (approved):
LGTM, apart from small nit noted.

PR #371, opened 2021-03-16
Add makefile fragment for INIT processing for macOS on x86_64

nbriggs:
We were missing the case for x86_64 macOS init processing.


rmkaplan:
Looks good

PR #372, opened 2021-03-16
Improvements to package/atom access in the sysout from maiko C code

nbriggs:
There were a number of sub-optimal behaviors in the package/atom handling code used in C to access atoms in the running image.  These changes make error reporting more explanatory and remove some unneeded functions.


masinter (commented):
I don't know how to test this. I doubt that it will work with "Fat" strings of 16 bit characters.

This is just stuff in URaid though, right?

nbriggs:
@masinter -- the S_TOPVAL and S_MAKEATOM were used from the debugger (dbx) and are no longer necessary.
The other changes are in code used by the rest of the maiko code during startup initialization -- they get native pointers to Lisp atoms in the loaded VM that are used for various purposes.  The new error handling will show up if you try to use a standard lde instead of ldeinit to do the step in the loadup which requires ldeinit.

masinter:
I'd like to get rid of ldeinit if possible, but that's another issue

PR #375, opened 2021-03-18
Command line: fix potential buffer overruns in argument/env variable handling

AbeJellinek:
Just started on this; not ready to merge yet, but I'm creating this so we have a tracking PR. The general goal here is to make `run-medley` redundant while keeping backwards compatibility in case users want to keep using it. The easy parts will be integrating the various small tweaks that it provides to make passing parameters easier. The hard part will be making lde use a search path for loadups so that it can work without `MEDLEYDIR`, which (while a challenge) is absolutely necessary before we can start packaging for `dpkg`/`apt`, Homebrew, and so on.


nbriggs:
OK -- but I'm already working in this code, an xrdopts.c is not the place to do it.  I have a PR in for replacing all the hand-crafted argument extraction with getopt (and friends).

nbriggs:
BTW -- this change on its own is OK, and sufficient to ask for a review and merge.  WaywardMonkeys and I have settled on a work style where PRs are very small so that they can be reviewed easily.

masinter:
MEDLEYDIR was a recent invention, and now's the time to rethink it. We need some way to refer to files in the repo or local installation without having to hard wire in the user's configuration.

AbeJellinek:
How about I submit this one change and then work on replacing MEDLEYDIR in a separate PR? As far as I understand it, that's not in the scope of what you're working on. I did not think that the MEDLEYDIR work would fit in xrdopts.c, but for command-line argument parsing in particular xrdopts either needs to be modified or replaced.

nbriggs (changes requested):
```
src/main.c:char sysout_name[MAXPATHLEN]; /* Set by read_Xoption, in the X version. */
src/main.c:#define MAXPATHLEN 128
```
It would be better to use `strlcpy` rather than `strncpy` which is also error prone.  See the man page for strlcpy/strlcat.  Use sizeof(sysout_name) for the size.

nbriggs (approved):
I disagree with the Linux people.  However, I'll approve this change since `strlcpy` is not in the C99 standard which we're trying to adhere to.

waywardmonkeys:
This got merged, but the result of the `strncpy` isn't guaranteed to be NULL terminated. And really, it isn't valid to truncate it either. If the length of the value is too long, it is better to report an error and exit, not carry on with an incorrect value.

nbriggs:
@waywardmonkeys -- agreed.  This is one of the failures that caused `strlcpy()` to be written.  If the final element of sysout_name were explicitly set to 0, along with moving at most length-1 characters, then we'd not risk an unterminated string, but we would have the incorrect name for the sysout -- which would either error with file-not-found, or open another file that happened to match the truncated name.

PR #376, opened 2021-03-19
Rename _curXXX structures to curXXX to avoid leading underscore

nbriggs:
It's not an issue in C, really, but checkers like to complain about leading underscores on identifiers


masinter (approved):
LGTM

PR #378, opened 2021-04-08
Cleanup USE_DPLI usage

nbriggs:
Until now, the only differentiation between using DLPI and using NIT for
the ethernet interface was expressed as a function of USE_DLPI.

This commit makes explicit when code is for the DLPI interface or the NIT
interface, with USE_DLPI and USE_NIT.  This is setup for using the BPF
interface to the ethernet as an alternative.

### DIFF
--- a/src/xrdopt.c
+++ b/src/xrdopt.c
@@ -19,9 +19,9 @@
 #include <unistd.h>
 #include <limits.h>
 #ifdef MAIKO_ENABLE_ETHERNET
-#ifndef USE_DLPI
+#if defined(USE_NIT)
 #include <net/nit.h> /* needed for Ethernet stuff below */
-#endif               /* USE_DLPI */
+#endif               /* USE_NIT */
 #endif               /* MAIKO_ENABLE_ETHERNET */
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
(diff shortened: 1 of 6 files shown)


src/xrdopt.c
```diff
@@ -297,9 +297,9 @@ void read_Xoption(int *argc, char *argv[])
...
     int b0, b1, b2, b3, b4, b5;
     (void)strncpy(tmp, value.addr, (int)value.size);
-#ifdef USE_DLPI
+#if defined(USE_DLPI)
     if (sscanf(tmp, "%d:%x:%x:%x:%x:%x:%x", &ether_fd, &b0, &b1, &b2, &b3, &b4, &b5) == 7)
-#else
+#elif defined(USE_NIT)
     if (sscanf(tmp, "%d:%x:%x:%x:%x:%x:%x:%s", &ether_fd, &b0, &b1, &b2, &b3, &b4, &b5,
                snit.snit_ifname) == 8)
 #endif /* USE_DLPI */
```
waywardmonkeys, line 305: Maybe edit this `#endif` comment?

waywardmonkeys (approved):
LGTM, minor nit.

src/xrdopt.c
nbriggs: Sure.  Thanks for noticing.

PR #382, opened 2021-05-16
Already there, OS-dependent variable SysReq

masinter:
in xc.c
```c
#ifdef XWINDOW
extern int Event_Req; /* != 0 when it's time to check X events
                                                 on machines that don't get them reliably
                                                 (e.g. Suns running OpenWindows) */
#endif                /* XWINDOW */
```
and only one place to patch.

### DIFF
--- a/src/keyevent.c
+++ b/src/keyevent.c
@@ -291,7 +291,7 @@ void getsignaldata(int sig)
   int i;
 
 #ifdef XWINDOW
-#if defined(sun)
+#if defined(sun) || defined(__CYGWIN__)
   if (Event_Req) {
     if (!XLocked++)
       getXsignaldata(currentdsp);


nbriggs (changes requested):
The code is incomplete -- what happens when it takes a timer interrupt when XLocked is true?

We're moving to "if defined(MAIKO_OS_CYGWIN)"

The description will show up in the log for all time -- how about a better description, and a reference to the issue that it is addressing?

nbriggs (commented):
How does this interact with KBDEventFlg, which seems to be performing a similar function?

masinter:
I don't know what happened (did I imagine it picked up type-in. Anyway, this isn't enough to make cygwin maiko work.

PR #383, opened 2021-05-18
Simplify OPFN/OPFNX macros by removing hand expansion of 1, 2, N bindings

nbriggs:
The OPFN and OPFNX macros had been coded with a hand expansion of a loop
pushing values on the stack to separately handle the cases of 1, 2, and
more values.  This actually lead to larger, likely slower, and certainly
less obvious code.


masinter (approved):
Simpler is better.

PR #385, opened 2021-06-24
Remove SIGPOLL usage

nbriggs:
Removes explicit usage of SIGPOLL by replacing indirect calls to getXsignaldata (via `kill(getpid(),SIGPOLL)`), when there was a deferred signal at XUNLOCK, by a direct call.  This requires the DspInterface be available from XUNLOCK usage sites, so some procedures and macros needed parameters adjusted.

There are also a few places where XLOCK/XUNLOCK calls are superfluous because the procedures they are in are always called with locks already held.


masinter (approved):
LGTM

PR #386, opened 2021-07-12
Dockerfile and github actions to build Maiko image

stumbo:
GitHub Action to build Maiko Docker Container on push to master.


stumbo:
Didn't realize there was already a build.yml file until I did my push.  Renamed my file to remove the collision.

.github/workflows/buildDocker.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # Setup some environment variables
+      - name: Prepare
+        id: prep
+        run: |
+          # Name of the Docker Image. 
+          DOCKER_IMAGE=${{ secrets.DOCKER_USERNAME }}/${GITHUB_REPOSITORY#*/}
+          # DOCKER_IMAGE=billstumbo/${GITHUB_REPOSITORY#*/}
```
stumbo, line 30: Delete commented line

.github/workflows/buildDocker.yml
```diff
@@ -0,0 +1,88 @@
...
+      # Checkout the branch  
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # Setup some environment variables
+      - name: Prepare
+        id: prep
+        run: |
+          # Name of the Docker Image. 
+          DOCKER_IMAGE=${{ secrets.DOCKER_USERNAME }}/${GITHUB_REPOSITORY#*/}
```
stumbo, line 29: This needs to change.  DOCKER_USERNAME isn't a direct mapping to the dockerhub repository.  Just use interlisp, no need for anything fancy.

nbriggs (commented):
How does one test this?

stumbo:
Thanks an excellent question - one that doesn't have a satisfactory answer.  I developed the code and ran it against my fork of Maiko.  Those builds were successful prior to putting together this pull request.  The last run before I made updates to the file for merging into the Interlisp Maiko repository is documented here:
   https://github.com/stumbo/maiko/actions/runs/1021567691

The changes since then were naming - pointing the dockerhub save to the Interlisp repository location, and integrating my github actions with those that already existed in Maiko.

To run it end to end within my fork requires merging those changes into my master and resetting my SECRETS to match the Interlisp repo SECRETS.  I thought about that, but pushing Docker images build off my fork into the Interlisp DockerHub repo seems like a bad idea.  

The risk of failure is small and the damage that a failure would cause is minor.  Failure would result in not creating a Maiko Docker image.

nbriggs:
Looking at the output from your run suggests that it's trying to pull a LOT more packages into the image than are necessary -- for example it doesn't need all of gcc-9, gcc-10, clang-10 -- one C compiler will do, and if there's consistency in the makefiles so that we use the same compiler for all the linux based builds...

masinter:
For "manual" testing I'd like to make sure it's possible to build a docker container with experiment-yield-subr, in order to test out not "pinning" the CPU when running in the cloud.  Making a docker container from a (draft) release would seem to be the ticket. 
Anyway, need this before the corresponding medley PR will work so I'll pull the trigger.

PR #388, opened 2021-08-10
Remove code for unsupported Sun display configurations

nbriggs:
Initial cleanup removing all code that was #ifdef'd for SUNDISPLAY.
Other SunWindows dependent code may also be removable in the future.


masinter (approved):
Got rid of a lot of bloat!

PR #390, opened 2021-08-25
add start-release and release-one for separate maiko/medley releases

masinter:
start-release starts the process and release-one sets up the rest.

This is all just a hint for GitHub action automation.


stumbo:
Would it make sense to do this as a GitHub action? Like the Docker container construction action we released last month?  The Maiko Docker container currently does the build of each release.  We could move that into a separate Create Release Action and extend its functionality to do the work you're doing locally and pushing to GitHub.  

The workflow could either be triggered manually or by some specific event (a push to master, needing to build a container, etc.)   

If not, I'd suggest putting the release build scripts into a new directory and segregate them from the rest of the Maiko code base.

nbriggs:
Is this the way we want it to report that `gh` isn't installed?

```
% bin/release-one 
bin/release-one: line 4: gh: command not found
run from maiko/bin
```

masinter:
this is a placeholder to show a proposal for what should be done in a GitHub action

I did want to have more kinds of releases (with yield, with gc debugging on) and wondering about release naming

nbriggs (commented):
What did you do in GitHub that caused it to want to re-add commits that are already in the master branch?

masinter:
I was trying to do a rebase. I don't know what the differences are between
```
git pull --rebase
git rebase master
git rebase origin/master
```

nbriggs:
The "git pull --rebase" is the easy one:

When you do a "git pull" you're asking it to fetch the remote information for the branch (unless you say otherwise, it will get it from the place the branch was created from -- in our case it's almost certain to be "origin"), and then combine that information into your local copy of the branch.   As the "git help pull" documentation says:  
> git pull runs git fetch with the given parameters and calls "git merge" to merge the retrieved branch heads into the current branch. With --rebase, it runs "git rebase" instead of "git merge." 

Most of the time a "git merge" (as done by "git pull") is going to resolve as a "fast-forward" because the current branch head exists in the history of the the branch you're merging.  If it's NOT, then a merge creates a new commit recording the parents of the two branches that contributed, while a rebase backs out your changes, does the fast-forward, then attempts to reapply your changes on top of the new state.

The variations of rebase (git rebase master, git rebase origin/master) are more complicated, and without re-examining the documentation I don't know that I can get the subtleties correct.  As far as I remember, it's almost never the right thing to do to "get rebase origin/master".   Doing a "git fetch" and then working with the local state of a branch will be more likely not to lead to confusion.

masinter:
is it possible to rewind this to aa389f1 then?

nbriggs:
I can fix it up and do the merge into maiko's master-- I assume that's:  Adding the start-release and release-one scripts, with a check that `gh` is installed before it tries to do anything, and no special color things in the output?

nbriggs:
The scripts do need the proper test for whether `gh` is installed -- running it to check the version when it isn't present just aborts the script and doesn't tell you you need to have the "gh" package installed.  By the way, the current release of gh "Require Go 1.16+ #4175 - drops support for OS X El Capitan" -- which means that the machine I normally compile on will never be able to use the "gh" tool.

masinter:
i rewound to commit  aa389f1 using git checkout  aa389f1, and then made a new branch and pr #395

nbriggs:
OK.  You can delete this branch then.

PR #392, opened 2021-09-01
Rewrite keyboard and async I/O handling to improve performance and reduce dependency on async I/O signals

nbriggs:
Replaces the SIGIO handler, which used to process X events in the interrupt context, with
a stub routine that sets a flag to signal processing is required.
Actual event processing is moved to the main dispatch loop where Lisp periodic interrupts
are handled.

Removes the X connection file descriptor from the set of fds contributing to SIGIO events
and moves the processing of X events to where the Lisp periodic interrupt is handled in
the main dispatch loop.  This code is already guarded by a check for XPending() so can
be called regardless of whether any file descriptors are known to be ready.

Actual processing of async I/O events and X events are handled by procedures
process_io_events() and process_Xevents() respectively.  For the most part these are
a renaming of getsignaldata() and getXsignaldata().

The Lisp periodic timer (VTALRM) was set to operate with a 25000 us period (40 Hz),
but on modern hardware it is possible to run this timer with a period of 10000 us (100 Hz)

Incidentally, a bug was noted (and fixed) in the X event handling code for motion events:
Mouse motion without any keyboard activity should not add an entry to the keyboard event
ring buffer as these events do not represent a key state change.
Since the ring buffer is of limited size, when it is filled new events are ignored
until the buffered events are processed.  This resulted in the loss of a key/mouse button
transition (up or down) if the mouse was moved about "too much" between keyboard events.

A few incidental cleanups were also made:
  - KBDEventFlg initialization fixed (wrong semantic type)
  - Event_Req renamed to IO_Signalled (more appropriate name)
  - int_io_open() sets up process (self) to handle SIGIO generated by O_ASYNC operations
  - LOCK_X_EVENTS turned off since X library calls can no longer happen in an interrupt context


masinter:
no joy on cygwin yet
image

nbriggs:
OK -- that code wouldn't be used with no async I/O set up -- given that cygwin doesn't even have the O_ASYNC symbol, I guess I can make that conditional on O_ASYNC being defined.

nbriggs:
> OK -- that code wouldn't be used with no async I/O set up -- given that cygwin doesn't even have the O_ASYNC symbol, I guess I can make that conditional on O_ASYNC being defined.

@masinter -- could you check if it happens to have FASYNC defined in one of the include files?

masinter:
no problems otherwise, the cygwin problem is old. It's just close -- I can get it to accept input sometimes by (in the terminal window) control-Z / bg / fg  interspersed with clicking in the medley window.

@nbriggs 

```
$ grep -r ASYNC /usr/include/sys/
/usr/include/sys/features.h:#define _POSIX_ASYNCHRONOUS_IO              1
/usr/include/sys/features.h:#define _POSIX_ASYNCHRONOUS_IO                      200809L
/usr/include/sys/mman.h:#define MS_ASYNC 1
/usr/include/sys/signal.h:#define SI_ASYNCIO 4    /* Indicates completion of asycnhronous IO */
/usr/include/sys/unistd.h:#define _SC_ASYNCHRONOUS_IO              21
/usr/include/sys/unistd.h:#define _PC_ASYNC_IO                      9
/usr/include/sys/_default_fcntl.h:#define       _FASYNC         0x0040  /* signal pgrp when data ready */
/usr/include/sys/_default_fcntl.h:#define       FASYNC          _FASYNC
/usr/include/sys/_default_fcntl.h:#define       F_GETOWN        5       /* Get owner - for ASYNC */
/usr/include/sys/_default_fcntl.h:#define       F_SETOWN        6       /* Set owner - for ASYNC */
```

masinter:
i went to try this out and found the change from 0 to SA_RESTART in PR didn't improve response to input under cygwin.

masinter:
@nbriggs it isn't asking for review but I'll let you press the Squash and merge button

nbriggs:
OK -- merged.  Do you want to try it on your mac and/or WSL2 setups and see whether it feels any different there?  On my mac it feels a little smoother, and can still respond to keys when it's doing heavy compute (or display) tasks.

PR #394, opened 2021-09-04
Replace uses of atoi() with strtol() or stroul() as appropriate

nbriggs:
As a side-effect of this change, we also resolve a a problem with
signed file version numbers, so that instead of version 2147483647
wrapping to -2147483648 we can go as far as 4294967296 before we
have issues.  Various sprintf() formats get changed from %d to %u.
The DOS version code is left behind as int versions.

Closes Interlisp/medley#455

### DIFF
--- a/inc/lispver1.h
+++ b/inc/lispver1.h
@@ -39,7 +39,7 @@
 	NumericStringP(vp, YES, NO);					\
        NO: *vp = 0;							\
        YES:								\
-	if ((*vp)) ver = atoi(vp);					\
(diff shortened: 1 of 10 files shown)


masinter (commented):
I started to review this, and got lost imagining negative version numbers for partial restore of git versions. (where {DSK}foo;-3 would be 3 versions back.
I don't suppose that was a possibility?

nbriggs:
nope... it never worked that way.  What happened was that if you have version 2147483647 the next version was -2147483648.

masinter (approved):

PR #398, opened 2021-09-14
add yield subr

nbriggs:
Provides a subr to do a `nanosleep()` call so that from the Lisp level you can have some control over how much of the CPU Medley is using.


nbriggs:
The changes are easier to review if you look with "ignore whitespace" (or on the command line, "git diff -w ...")

masinter (approved):
Great, thanks

PR #399, opened 2021-09-29
Implement -noscroll option to ldex to disable X window scroll bars

nbriggs:
Adds a -noscroll option, parsed as an X option, also accessible via
resource ldex*noscroll, which avoids adding the bottom and side scroll
bars and the bit-gravity control buttons to the main Lisp display window.

Unless the geometry given for the X window in which the Lisp screen is
displayed is at least as big as the Lisp screen part of the Lisp screen
(bottom, right) will not be visible.

At least on macOS with XQuartz, maximizing the X window will bring it to
the size of the Lisp screen (or the size of the display, whichever is smaller)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -210,7 +210,8 @@ int UnixPipeIn;
 int UnixPipeOut;
 int UnixPID;
 int please_fork = 1;
-
+/* disable X11 scroll bars if requested */
+int noscroll = 0;
 /*** STACK handle staff(Takeshi) **/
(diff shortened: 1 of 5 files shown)


nbriggs:
To be maximally useful, this will require changes to the run-medley script, since that script isn't prepared to pass on unknown options to lde/ldex.

masinter (approved):
lgtm, more involved than I had imagined

PR #402, opened 2021-10-10
The pointers Atomspace, Spospspace, Snegspace, Arrayspace can be removed

nbriggs:
These pointers are neither initialized nor referenced anywhere in the code.


masinter:
fixed https://github.com/Interlisp/medley/issues/519

PR #403, opened 2021-10-20
[WIP] Add prototypical SDL support.

ecraven:
This is just a Merge Request to discuss, it is not ready to be merged in any way.

UPDATE: read along, all of the following is not true any longer ;)

Build with cmake. This will create a new backend (ldesdl).
- Resolution can only be set by editing the variables in sdl.c (UPDATE: edit the `init_SDL` call in main.c)
- Key repeat does not work.
- Still problems with keysyms that implicitly contain modifiers.
- The entire screen is bitblted onto the SDL display every frame.


ecraven:
@nbriggs I think with the last push it should now only update the damaged parts. I didn't look into expose events in SDL yet, not sure how well those are supported.
If you have a bit of time, could you maybe try whether this works for you on a mac and even possibly on windows?

nbriggs:
I can try on MacOS, assuming I can install/compile SDL without having to use "brew" (or anything like it).   I don't have a Windows box.

ecraven:
> I can try on MacOS, assuming I can install/compile SDL without having to use "brew" (or anything like it). I don't have a Windows box.

That would be great! Please take care to use SDL2, not any of the older versions ;)

nbriggs:
The includes in sdl.c for SDL files should be just `"SDL.h"` and `"SDL_keycode.h"` -- the `SDL/` is deprecated (according to the CMake people) because FreeBSD doesn't put these includes in an SDL subdirectory.

ecraven:
> Yeah, I've been following along.  It works much better now.   `FB
> *.*;*` in the Lispusers directory is a good test for how it's doing on
> single character BLTCHAR, and that seems to be acceptably fast now.

Playing around with cygwin, not much success so far ;)

So, do you think this SDL thing is something worse pursuing further?
Should I clean up the code, to maybe start on an actual real Merge
Request?

Thanks for the help!
Greetings, Peter

nbriggs:
If we can sort out the input (and things such as doing the 3-button mouse emulation) it's probably worthwhile pursuing.
When I do a bunch of FB redisplays, over a period of 35s runtime, the `sdl_bitblt_to_screen()` is responsible for 36% of the execution time -- so it's still more than it should be.  Under X11, `clipping_Xbitblt()` which is the equivalent, is too small to show up.

ecraven:
I think I already added and pushed that, ran across that too.. so, what's the best way to proceed here? I'm no windows programmer, but I can try to fix things up.
However, this is mostly unrelated to SDL, so maybe we should try to separate those two changesets.
Should we discuss the next steps on monday?

nbriggs:
I think a Windows change set is orthogonal to SDL, though enabled by it.
The last change you made to the CMake configuration stopped it from compiling on macOS
```
[  2%] Linking C executable lde
ld: library not found for -lSDL2
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
I think I need to make more changes in the FindSDL2.cmake.  I believe all the variables it creates have to be SDL2_... instead of SDL_...

Yes, we should talk about it on Monday. I'll continue to poke at getting things working properly on the mac between now and then, though.

ecraven:
Sorry, it's the `SDL2::SDL2` vs. `SDL2` alone thing.. for windows, only `SDL2` alone worked, on linux I need `SDL2::SDL2` too.. I need to learn more cmake
If there's anything I can help with to make things easier for you, please tell me. If that seems like a reasonable approach, I'll let you work on MacOS stuff, and not touch things too much for now. If it works satisfactorily on MacOS too, I'll try to clean up this entire changeset.
What are the things that are still missing, apart from the following?

- 3 button emulation: is there code for this? how *should* it work?
- command line parameters: I'll just add three parameters, width, height and pixelscale, ok?
- input: as the "problem" only appears in my very specific setup, I don't think this will be an actual problem for others.. does keyboard and mouse input work ok for you? any glitches there?

nbriggs:
These two changes broke compilation on macOS and FreeBSD --

```
-  SET(MAIKO_DISPLAY_SDL_LIBRARIES SDL2::SDL2)
+  SET(MAIKO_DISPLAY_SDL_LIBRARIES SDL2)

-  TARGET_LINK_LIBRARIES(lde SDL2::SDL2)
+  TARGET_LINK_LIBRARIES(lde SDL2)
```

nbriggs:
- 3 button emulation: is there code for this? how should it work?
On the mac, XQuartz does this outside of any of our code. One can configure (for example) rCommand and rOption to make the mouse left button click come through as middle or right button.   Usually you'd just have a 3-button mouse, but a mac trackpad or "normal" mouse has only one button (but can be configured for right button on two-finger click)
- command line parameters: I'll just add three parameters, width, height and pixelscale, ok?
That would work for a few days.  It should match the existing options for things like `-t "window title"` and `-m memorysize` and the same `-sc WxH` syntax for window size, otherwise we end up having to special case everything in the run-medley script for no good reason.
- input...
There seem to be some glitches, but I can't characterize them yet.  I need to systematically explore.  It will be easier when I have all the mouse buttons.

nbriggs:
Well all the references to X11 libraries are X11::X11 so with no understanding of what it's really doing I'm not surprised that the SDL2 libraries are SDL2::SDL2

nbriggs:
```
./CMakeCache.txt:SDL_INCLUDE_DIR:PATH=/Library/Frameworks/SDL2.framework/Headers
```
and 
```
% ls /Library/Frameworks/SDL2.framework/Headers
SDL.h				SDL_endian.h    [...]
```
there is no subdirectory SDL or SDL2.
But... I don't see where SDL_INCLUDE_PATH would get used, though I also think that I need to fix my FindSDL2.cmake file so that all the variables it creates are actually SDL2_... instead of SDL_... since they are *supposed* to be prefixed with the package.

nbriggs:
Handle repeating keys
```
diff --git a/src/sdl.c b/src/sdl.c
index bbff60b..73dc365 100644
--- a/src/sdl.c
+++ b/src/sdl.c
@@ -347,6 +347,10 @@ void process_SDLevents() {
       break;
     case SDL_KEYDOWN:
       printf("dn ts: %x, type: %x, state: %x, repeat: %x, scancode: %x, sym: %x <%s>, mod: %x\n", event.key.timestamp, event.key.type, event.key.state, event.key.repeat, event.key.keysym.scancode, event.key.keysym.sym, SDL_GetKeyName(event.key.keysym.sym), event.key.keysym.mod);
+      if (event.key.repeat) {
+        /* Lisp needs to see the UP transition before the DOWN transition */
+        handle_keyup(event.key.keysym.sym, event.key.keysym.mod);
+      }
       handle_keydown(event.key.keysym.sym, event.key.keysym.mod);
       break;
     case SDL_KEYUP:
```

ecraven:
Also another question, do we want to keep using makeright, or will we switch to cmake fully? I've only worked on cmake now, I haven't touched makeright and its files at all..

ecraven:
> I've filed a bug report with the SDL folks for the missing FindSDL2.cmake (or whatever they think is appropriate for CMake to be able to find everything right out of the box on a mac, or with at most an environment variable setting)

Great, thanks!

ecraven:
here's ASDF:
```dn ts: 1eaf, type: 300, state: 1, repeat: 0, scancode: 4, sym: 61 <A>, mod: 0
dn A -> lisp keycode 21 (0x0)
dn ts: 1eb0, type: 300, state: 1, repeat: 1, scancode: 4, sym: 61 <A>, mod: 0
up A -> lisp keycode 21 (0x0)
dn A -> lisp keycode 21 (0x0)
other event type: 771
up ts: 1f0d, type: 301, state: 0, repeat: 0, scancode: 4, sym: 61 <A>, mod: 0
up A -> lisp keycode 21 (0x0)
dn ts: 202b, type: 300, state: 1, repeat: 0, scancode: 16, sym: 73 <S>, mod: 0
dn S -> lisp keycode 20 (0x0)
dn ts: 202c, type: 300, state: 1, repeat: 1, scancode: 16, sym: 73 <S>, mod: 0
up S -> lisp keycode 20 (0x0)
dn S -> lisp keycode 20 (0x0)
other event type: 771
up ts: 2089, type: 301, state: 0, repeat: 0, scancode: 16, sym: 73 <S>, mod: 0
up S -> lisp keycode 20 (0x0)
dn ts: 22eb, type: 300, state: 1, repeat: 0, scancode: 7, sym: 64 <D>, mod: 0
dn D -> lisp keycode 5 (0x0)
dn ts: 22eb, type: 300, state: 1, repeat: 1, scancode: 7, sym: 64 <D>, mod: 0
up D -> lisp keycode 5 (0x0)
dn D -> lisp keycode 5 (0x0)
other event type: 771
up ts: 2344, type: 301, state: 0, repeat: 0, scancode: 7, sym: 64 <D>, mod: 0
up D -> lisp keycode 5 (0x0)
dn ts: 244c, type: 300, state: 1, repeat: 0, scancode: 9, sym: 66 <F>, mod: 0
dn F -> lisp keycode 35 (0x0)
dn ts: 244c, type: 300, state: 1, repeat: 1, scancode: 9, sym: 66 <F>, mod: 0
up F -> lisp keycode 35 (0x0)
dn F -> lisp keycode 35 (0x0)
other event type: 771
up ts: 24a3, type: 301, state: 0, repeat: 0, scancode: 9, sym: 66 <F>, mod: 0
up F -> lisp keycode 35 (0x0)
```
not sure why, but I do get one repeat for every keypress in SDL (but not with evtest or xev) :-/ might be a local problem on my machine, I'll check on another machine in the next few days

nbriggs:
I added
```
+    case SDL_MOUSEWHEEL:
+      printf("mouse wheel\n");
+      break;
```
to the events, and will look at what it's telling us for the track-pad finger movements on the mac (and presumably for people with an actual mouse wheel).

nbriggs:
```
   sdl_texture = SDL_CreateTexture(sdl_renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STREAMING, width, height);
-  printf("-> %x\n", sdl_texture);
+  printf("-> %p\n", sdl_texture);
   buffer = malloc(width * height * sizeof(Uint32));
```

nbriggs:
Deciding that I wanted a version of SDL with debug symbols to see what on earth was going on, I picked up the current source (2.0.17) and built it myself.

Having done that, the timing has changed significantly -- it's much more reasonable now:
SDL(texture): 1.91s cpu/minute elapsed, output 1380 ms (837 in bitblt_to_screen), input 28ms
SDL(blt): 0.569s cpu/minute elapsed, output 48.6ms, input 26 ms

The big difference between the blt output and the texture output is that with the blt output I only ever convert the damaged region from 1 bpp to 32 bpp and then SDL_BlitSurface that region to the window surface.   The blt numbers would be much more like the texture numbers if more of the screen were being updated.

ecraven:
I've rebased this onto current master, built it (still seems to work), also pushed my non-finished SDL Cursor work (the cursor works fine, but isn't cached yet, should just use the same mechanism as X uses right now).
@nbriggs I've pushed it to upstream/sdl, so that we can collaborate on it. Sorry for not having the time yet to clean it up more :-/

nbriggs:
Built SDL on Solaris 11.4/SPARC, after an issue with the SDL install (doesn't install all the include files that it should), I can build `ldesdl` without any trouble.   SDL on Solaris uses the X11 library backend and attempts to bring up the remote display on my Mac, but chokes in creating the renderer --
```
Creating renderer...
ldesdl          -> libSDL2-2.0.so.0:*SDL_CreateRenderer(0x1001f4db0, 0xffffffffffffffff, 0x2)
ldesdl          ->       libc.so.1:*malloc(0x110, 0x0, 0x0)
ldesdl          ->       libc.so.1:*free(0x1002fb340, 0x0, 0xffffffff7e92eb80)
ldesdl          ->       libc.so.1:*calloc(0x1, 0xa0, 0xffffffff7e7deef0)
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server "localhost:10.0"
      after 130 requests (130 known processed) with 41 events remaining.
```
So I'll be seeing if it gets further with the bitmap version rather than the texture version.

ecraven:
@nbriggs great! I've played around with the renderer some more, but I don't think this is getting much faster, without actually using full OpenGL, which I'd like to avoid... Let me know if I can help with testing your approach!

nbriggs:
I've incorporated the bitmap surface changes -- it's controlled by defining the preprocessor symbol `SDLRENDERING` -- if it's defined it uses your original code (slightly refactored, and a couple of issues addressed) but if not defined it uses `SDL_BlitSurface` (or `SDL_BlitScaled` if necessary) to the window's surface.

Compiling/running on Solaris, it *tries* to work, but manages to put up a bright yellow and black bitmap that's pretty much scrambled over an X connection via ssh.   There's some work to be done there, and even then it may be totally unusable speed-wise.

I haven't done performance measurements yet.

ecraven:
@nbriggs I've tried it, but the code (strangely) only displays entirely black... still investigating what's wrong here, I think it's related to the RGB332 pixelformat.. 
I've tried to run FB and noticed that the initial display seems fast enough, but Shaping the window is abysmally slow.. any ideas as to why this is?
From looking at the damage notifications, it seems that each character is drawn one by one, and we get two damage notifications per character?

nbriggs:
@ecraven -- I'll look at the FB speed, I suspect it's mostly in enumerating the files (again) when its reshaping, and it handles it differently the first time around.

I compiled on FreeBSD 13.0, and over an SSH connection it used X back to my mac, with a correct display, but too slow to do much.

You'll have noticed that I used the same pixel format that you did.  I'll tidy up the allocation of the intermediate display buffer so that it's easier to change the pixel size and still get the buffer the right size.

nbriggs:
@ecraven  -- While looking at the buffer related things I also looked at what textures the renderer supports (on a mac), and it says it's an "opengl" renderer, with support for textures
```
"SDL_PIXELFORMAT_ARGB8888"
"SDL_PIXELFORMAT_ABGR8888"
"SDL_PIXELFORMAT_RGB888"
"SDL_PIXELFORMAT_BGR888"
"SDL_PIXELFORMAT_YV12"
"SDL_PIXELFORMAT_IYUV"
"SDL_PIXELFORMAT_NV12"
"SDL_PIXELFORMAT_NV21"
"SDL_PIXELFORMAT_UYVY"
```
so I guess choosing one of those would result in less conversion overhead?  I'm setting up a version which picks up the first format returned and then asks for the black & white pixel values for that format, and the width of a pixel (in that format) in order to create the intermediate buffer.  That way it should adapt a little to the underlying machine.

nbriggs:
@ecraven  -- Was there a reason not to do SDL_TextureLock()..SDL_TextureUnlock() and update from the Lisp screen bitmap directly into the texture pixels?  This avoids having any intermediate buffer, as far as I can see.  You have to get the locking region exactly correct because anything that is locked MUST be updated.
I've got an almost working version that does this (and am tracking down the "almost")

ecraven:
@nbriggs just my lack of knowledge ;D thanks for working on this!

nbriggs:
The SDL documentation does not give the clearest examples... I ran across the LockTexture stuff by accident as I was looking for something else, and it suggested that if you were using streaming textures you should lock and update the texture pixels directly.   I'll push a new version, possibly as a separate branch, as soon as I've got something that gets the displayed bitmap correct, even if it's not the cleanest!

nbriggs:
@ecraven -- I've pushed a branch named sdl-texture-direct, started from the current sdl branch.
Totally messy.  Defaults to the SDLRENDERING version.  In the case of streaming textures, I don't understand the function of SDL_RenderCopy & SDL_RenderPresent -- but they seem to be necessary.

With these changes, at least on my Mac, an idling Medley which is just flashing the cursor takes very little CPU time (also has background-yield loaded, though). 

Compile it after configuring with `cmake .. -DCMAKE_BUILD_TYPE=Release` if you want to see the actual speed, since `...=Debug` will have it spend all its time in sdl_bitblt_to_texture() twiddling bits.

nbriggs:
Things to think about:
    - cursors need to honor the pixelscale if we're going to have pixelscale
    - setting the cursor position needs to clip to screen size
    - access to the cursor bitmap needs to account for host byte ordering with the (predefined) access macros, as with the access to the display memory, GETBASEWORD and friends

nbriggs:
@ecraven -- Looking at the frame-rate: the `process_SDLevents()` call, driving the display update, is from the part of the emulator code where the Lisp periodic interrupt handling happens.  I trigger the Lisp interrupt checks off the ITIMER_VIRTUAL which is set at 10 ms of process execution time. There are a very few situations where it could reenter this code in under 10 ms (experimentally, about 0.2% of cases), so I agree with your comment that we could remove the timing check.

nbriggs:
How about this for situations where we have both SDL and X11 available -- in ldeboot.c if both SDL and XWINDOW are defined accept the "-d"/"-display"/DISPLAY environment variable, and if it's "SDL|sdl" then try to start ldesdl, otherwise pull out the display name and check it as it currently does and if the X connection succeeds start ldex.  If only one is defined there's no choice, just do it.

lassik:
> How about this for situations where we have both SDL and X11 available

Nowadays there exist at least two alternative X environment variables, `DISPLAY` and `WAYLAND_DISPLAY`. If you want both in the same executable, it's best to have a separate command line flag like `--display sdl` or `--display x`.

nbriggs:
```
OPTIONS
       Most X programs attempt to use the same names for command line  options
       and  arguments.  All applications written with the X Toolkit Intrinsics
       automatically accept the following options:

       -display display
               This option specifies the name of the X server to use.
```
this would preclude using `-display x`, since in the direct to X11 case we use the X toolkit to parse the command line for all the other X related options.  I looked at the SDL source and we'd be OK with `-display sdl` since it passes NULL to `XOpenDisplay` and (from XOpenDisplay doc) "On a POSIX-conformant system, if the display_name is NULL, it defaults to the value of the DISPLAY environment variable." (that is, nobody is helpfully parsing any command line args behind the scenes).

lassik:
Good point.

Or alternative names:

* `-backend sdl`
* `-ui sdl`

masinter:
If we're considering changing the structure of lde, ldex, command line invoaction etc. in order to account for a new backend, now that there's a PoC for this, let's take a moment to consider the context.  The scripts around run-medley and lde and etc. were all pretty ad hoc.  I thought I heard there was some possibility of having just one executable instead of two. And changing the name from "lde" to "medley", I think.   And doing a better job of invoking pixel scaling ... And...

nbriggs:
> possibility of having just one executable instead of two.
This makes it a bit more difficult to have a setuid root executable that can open the network device and then drop privs and fork the rest of the emulator.  People would be rightly suspicious of depending on a larger piece of code that may or may not correctly drop (and not attempt to regain) root privs.

However, if we think that it's possible on all the systems we care about to assign network access capability to either the user or the program, without giving root access, it's certainly worth considering.

There's also the possibility of getting rid of the initial fork/exec and pipe communication for the external processes -- `CREATE-PROCESS-STREAM` and the like, if `posix_spawn()` is available on all the systems we care about (is it available on Cygwin? Windows?)

nbriggs:
Updated the sdl-texture-direct branch with some cursor cleanups.
Now moves the cursor up the cursor linked list only if it's past the first two items -- Medley tends to flip back-and-forth so there's no reason to reorder if it was there because it will probably go back shortly.  Fixes up type usage to match emulator types and accounts for host byte order when accessing the emulator cursor bytes.

masinter:
@nbriggs cygwin` isn't important if we have Windows/SDL. Whether Windows implements copy-on-write isn't clear. (Interlilsp-10 on Tenex had copy-on-write).

https://news.ycombinator.com/item?id=19621973 
https://news.ycombinator.com/item?id=18073906

lassik:
> This makes it a bit more difficult to have a setuid root executable that can open the network device and then drop privs and fork the rest of the emulator.

Setuid is perhaps the biggest anti-pattern in Unix programming - it should only be used for a handful of system programs like `su` and `sudo`. Applications should never use it.

If the ethernet device needs to be opened, it would probably be best for the sysadmin to create a group for users who need access to the device, then `chown` the device file such that the group owns it. Would this work?

If that doesn't work, there should be a minimally small daemon program that runs as root (but is not setuid) and gives other programs access to the device.

nbriggs:
>If the ethernet device needs to be opened, it would probably be best for the sysadmin to create a group for users who need access to the device, then chown the device file such that the group owns it. Would this work?

That's the second choice if your system lacks a security capability architecture such as Solaris' [role-based access control](https://docs.oracle.com/cd/E23824_01/html/821-1456/rbac-1.html).

If you install Wireshark on a Mac, using libpcap,  it does the group setup etc. for you -- installing `/Library/Application Support/Wireshark/ChmodBPF/ChmodBPF` that is run by launchd to create and set group access on the /dev/bpfNNN devices to group access_bpf.  I haven't investigated whether ChmodBPF was distributed with libpcap or wireshark.

Related stuff:  We've used the DLPI and NIT interfaces in the past, but if we're going to do raw ethernet access now then libpcap/BPF is probably the way to go. I think that BPF is now available on all systems where DLPI/NIT are (or were), so we can simplify the code.

lassik:
Neat! You know much more than I do about this stuff. libpcap is indeed popular nowadays. Whatever Wireshark does is probably a reasonable starting point.

nbriggs:
Updated the sdl-texture-direct branch with the code to fully deal with cursors in pixel-scaled presentation.

orcmid:
I do mostly desktop Windows, and the amount of tacit knowledge there is about Maiko and Medley melts my brain.

I am interested in graphic systems for gaming and took a deeper look at SDL2 because of this conversation.  I have run into much chatter about raylib/SDL2/OpenGL/Vulkan and have been sticking with raylib for my own education at this point.  My way ahead is by reverse-engineering sources (and CMakes) down to where I can do it all with command-line builds, getting the individual modules to compile, and then attempt integrated builds of .exe files.  (I am looking to see how I can understand maiko that way, using procedures I am using elsewhere anyhow.)

With regard to SDL2, [this](https://github.com/libsdl-org/SDL/tree/main/docs) is what put me off. 

> The best way to learn how to use SDL is to check out the header files in the "include" subdirectory and the programs in the "test" subdirectory. The header files and test programs are well commented and always up to date.

I don't believe that statement.  The README for Windows is out of date and also peculiar concerning Visual Studio version dependencies.  I followed an #include chain looking for where things like `SDL_NO_RETURN` and a bunch of others are introduced and got lost, the so-called well-commented being elusive.   I also saw somewhere that SDL2 builds as a DLL only; t can't remember where I saw it.  There are some consistencies where `SDL_Mumbles` are used and C/C++ types are used.  Sometimes specfic-size number types, other times not. Much to tease out.

I appreciate that configuring for different platforms is tricky.  I was never very happy about how that worked when Unix flavors were the platform everywhere that EBCDIC wasn't.  It's more difficult now, and having code uncluttered by the platform dependencies everywhere is seriously tricky. 

I don't have an answer.  I just want to complain and prey that `#ifdef` and `#elif' not be the answer running through all the code.  (This from a guy who only does release builds too :)).  It is regretable that SDL2 does not smell better.

masinter:
*bump* will recent changes to maiko make this harder or easier to complete?
Were we waiting for a bugfix from SDL?
Do @devhawala 's dodo changes reduce the need for root or special group access?

nbriggs:
It means there is a lot of editing to do.   We were waiting for the SDL people to fix an integration with CMake issue, they may have got around to releasing that, I haven't checked recently as they were moving very slowly.  Re: dodo changes -- not sure, I haven't read the code.

devhawala:
The Nethub extension to Maiko does not need any root/setuid/setgid priviledges: connecting to the Dodo Nethub uses plain TCP/IP client system calls.

The purpose of the Nethub is to avoid any special APIs or executable rights if possible. The only exception may be the NethubGateway which is a Nethub-client connecting to a PCap device, allowing client machines (real or emulations) to connect to Dodo XNS services (see [Dodo - Topology](https://github.com/devhawala/dodo#topology)), this one may need root rights to use the PCap device (at least on Linux).
But this is not the case for Maiko with Nethub support, as Maiko connects "natively" (directly through TCP/IP) to the Nethub.

BTW: some months ago i stumbled over the [Nano-X Window System](https://github.com/ghaerr/microwindows), which allows programs using one out of several UI-APIs to have their UI running on many other platforms (i did'nt have time to play with it yet, so no idea if it really works or which requirements must really be fulfilled)

nbriggs:
I have successfully rebased the interlisp/maiko SDL branch onto the most recent master and rebuilt with current SDL2 (2.24.2) on the Mac.

Required a fix to the SDL2 installation because the installed framework include files refer to an SDL2 subdirectory (`#include <SDL2/SDL_...>`), but the the include files are all at the same level in `/Library/Frameworks/SDL2.framework/Headers`.  Temporary fix is to `ln -s . SDL2` in the Headers directory.  Problem reported as libsdl-org/SDL#6513

masinter:
I wanted to try this on Windows, but I was left pretty confused about which SDL configuration to try.
Do we still need cygwin even if we don't need X?
Which config of SDL should I start with, out of the many available in 
https://github.com/libsdl-org/SDL/releases/tag/release-2.26.1

nbriggs:
Maiko depends on POSIX system calls for things as well as needing to talk to the keyboard/mouse/display.
 * MINGW lets you use gcc to compile programs that use Windows APIs (not POSIX) so that's not obviously what we want, unless the Windows APIs are source compatible with the POSIX APIs (and some of them might be)
 * Cygwin gives you a runtime library that translates between the POSIX and Windows APIs, so we'd probably want that
 * SDL used to be built/packaged by someone for Cygwin, but they appear to have abandoned that in 2015 or so
 * So you'll need to install the SDL2  source and compile it in a cygwin environment and then use the library that that produces to link with Maiko (plus installing the include files where a Maiko compile will find them) and SDL will presumably use some graphics output and keyboard input that I don't have any way of identifying.

orcmid:
> @nbriggs Maiko depends on POSIX system calls for things as well as needing to talk to the keyboard/mouse/display.

I don't understand enough about connecting the dots.  I do wonder if using the Windows Subsystem for Linux (WSL) would give you a more direct level of POSIX substrate.  SDL should work in that case.

A farther reach: The Standard C libraries are supported for native Windows builds. That may not be enough of POSIX and there is the question about SDL atop it, including how display loops and whatnot are accomplished at that level.  I suspect there might be a show-stopper or simply too much work.  I didn't check to know what the dependencies on Windows APIs might be presumed by the SDL distributions for Windows.

nbriggs:
SDL code based on this was merged into master.  We can close this WIP draft PR.

PR #405, opened 2021-10-31
Correct integer type warnings

nbriggs:
Corrects a number of issues related to casting values to a smaller type.

Warnings revealed usage of the wrong `TIMEOUT` (vs `TIMEOUT0`) macro, now corrected.
Some pointer arithmetic done with integers corrected to use modern `ptrdiff_t` type.
Comment added to (currently unused) code noting that you can't get a native address as a FIXP on a 64-bit pointer architecture.  If this code eventually gets used it will need to create/accept a 64-bit lisp object.

### DIFF
--- a/inc/timeout.h
+++ b/inc/timeout.h
@@ -61,8 +61,8 @@ extern	int	TIMEOUT_TIME;
 /************************************************************************/
 
 #define INTRSAFE(exp)				\
-  do {} while ((int)(exp) == -1 && errno == EINTR)
+  do {errno = 0; } while ((exp) == -1 && errno == EINTR)
 
 #define INTRSAFE0(exp)				\
-  do {} while ((int)(exp) == 0 && errno == EINTR)
+  do {errno = 0; } while ((exp) == NULL && errno == EINTR)
 #endif /* TIMEOUT_H */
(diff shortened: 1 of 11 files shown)


masinter:
We already have arbitrary precision integers (bignums). Not sure how to review these changes.

nbriggs:
> We already have arbitrary precision integers (bignums). Not sure how to review these changes.

bignums would work, though they are perhaps overkill since this code was likely never used for objects on which one would do arithmetic.  I'm guessing that these particular subrs will end up being removed, which is why I didn't build the C code to either create/access bignums or switch to returning, say, a CONS of a couple of FIXPs or an untyped memory hunk.

The only way to review the code is to look at the changes and the usage, I think (other than "does it compile", with or without warnings, before/after the change)

X_init is only ever called with 0 as the lisp bitmap, but if it weren't, it would be a Lisp-space pointer, as evidenced by the code that uses it:
```
  dsp->ScreenBitmap.data = (char *)(lispbitmap ? Addr68k_from_LADDR((LispPTR)lispbitmap) : 0);
```
The alternative would have been not to do the conversion here and leave it as a `char *`.  There's no real argument one way or the other, I think, OR we could remove the parameter entirely and always initialize to NULL -- but that would be a subject for review if you want to look at this and any other display xxx_inits that are present.

On INTRSAFE, TIMEOUT and their ...0 variants -- reading the spec for system and library calls that are passed to them, noting whether or not they set `errno` when they complete normally, and whether or not the call results in -1 error code or a (null) pointer on failure.

... and so on.

masinter (approved):
LGTM looked over edits and ran on it for a day

PR #406, opened 2021-11-05
Fix "warning: incompatible pointer types" in EVAL implementation

nbriggs:
nnewframe() as called from the EVAL implementation expects to be passed
the address of an array of two DLwords.  Do that, and compose the 32-bit
result correctly rather than passing the address of a (32-bit) LispPTR
and then having to swapx() that result to get the expected value.


masinter (approved):
nice

PR #407, opened 2021-11-05
Correct error for missing byteswap macros in tosfns.h EVAL implementation

nbriggs:
Commit c46fcce307018df6554805050ceb634d72e737c9 fixed a warning for
incompatible pointers to nnewframe() but did not consider that the
order of the DLwords is swapped depending on the endianness of the
system it is running on.  Add the necessary GETBASEWORD() macros to
access the items when constructing the pointer.

As written previously, it worked for little-endian but not big-endian systems.


nbriggs:
To be clear: this is how the value is stored into the achain words --
```
      GETBASEWORD(achain, 1) = result >> 16;
      GETBASEWORD(achain, 0) = result & 0xFFFF;
```
so when reconstructing the result from the `scratch` array passed in via `achain`, we need to use `GETBASEWORD()` to access the parts, rather than going directly via the array index.

PR #408, opened 2021-11-07
Fix "error: implicit declaration of function" from dspsubrs

nbriggs:
Fix "error: implicit declaration of function" from dspsubrs
    
Creates an sdldefs.h, declares a few more functions in sdl.c static, and includes sdldefs.h in dspsubrs.c when necessary.


ecraven (commented):
Thanks, looking good!

PR #410, opened 2021-11-23
Remove makepathname()

nbriggs:
makepathname() is only ever called with one of two constant string
arguments, in one case (DOS) such that it only does a strncpy(...),
and the other case (non-DOS) expanding "~" to the current user's home
directory. Additional code duplicates realpath() functionality but is
never used.

Closes interlisp/medley#578


masinter (approved):
looks good, as far as it goes; I have some feature requests that I'll make new issues for.

PR #411, opened 2021-11-30
Convert	"mkvdate" into a shell script instead of compiled C code

nbriggs:
For all systems other than DOS, use a simple shell script to create
the vdate.c (contains version date) that goes into every build.


masinter (approved):
👍

PR #412, opened 2021-12-03
Add some support for cross-compilation to MC68000 AmigaOS

nbriggs:
Add m68k to recognized CPUs (machinetype)
Add amigaos to recognized operating systems (osversion)
Add ability to override "config.guess" output w/ LDEARCH= to osversion
Add platform.h clauses to recognize AmigaOS 3 using gcc to cross-compile

### DIFF
--- a/bin/machinetype
+++ b/bin/machinetype
@@ -12,6 +12,7 @@
 os=${LDEARCH:-`./config.guess`}
 # o/s switch block
 case "$os" in
+        m68k-*)		      echo m68k ;;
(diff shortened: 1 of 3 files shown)


nbriggs:
This shouldn't affect any other systems we already compile for.  It does provide a mechanism for overriding the output of "config.guess" in the value that is provided to both bin/osversion and bin/machinetype, but new experiments would still need to add entries to these scripts if either the CPU type or the OS isn't already recognized.  This adds m68k-*-amigaos to the recognized combinations.

PR #414, opened 2022-01-22
Remove support for GET-NATIVE-ADDR-FROM-LISP-PTR and GET-LISP-PTR-FROM-NATIVE-ADDR

nbriggs:
Since the native addresses for objects within the Lisp memory may be at locations
that can't be represented in 32-bits conversion between a native address and a
Lisp FIXP can't be supported.  For now, return NIL for conversion in either direction.

There are no known users of these two subrs.  Should it be necessary to reimplement them
at some future time the representation will need to be something that can hold 64 bits.


masinter (approved):
LGTM

PR #415, opened 2022-02-11
Fgh004: Adding new workflow that builds maiko releases and docker images

fghalasz:
Adding new workflow that builds maiko releases for Linux/amd64, Linux/arm64, Linux/arm7, Macos/amd64. (Not Macos/arm64 as yet.) Also builds and pushes to DockerHub docker images for Linux/amd64 and Linux/arm64.

Can be called as a reusable workflow or dispatched from the Github Actions page. 

Delete old buildDocker.yml workflow.

Has not yet been tested on Interlisp/maiko - only on fghalasz/maiko fork because can't test manual dispatch workflows off the main (default) branch.


masinter:
don't need to add source zip and tgz because GitHub already adds them

fghalasz:
Fixed in pull request for FGH008 (PR#418)

PR #417, opened 2022-02-12
Fgh008: Taking source builds out of buildReleaseInclDocker workflow

fghalasz:
Taking source builds out of buildReleaseInclDocker workflow as per Larry's comment on PR#415.  Github does source build automatically when you publish a release.


fghalasz:
Will resubmit after rebasing to avoid merge conflicts.

PR #420, opened 2022-03-29
Correct usage of stat.st_mode result for testing S_IFREG and S_IFDIR

nbriggs:
stat.st_mode cannot be tested for whether a node is S_IFDIR with
    if (stat.st_mode & S_IFDIR) ...
since S_IFDIR bit (0040000) is a subset of S_IFSOCK (0140000) bits.
A correct check is
    if ((stat.st_mode & S_IFMT) == S_IFDIR) ...
or alternatively, since the convenience macros are defined on all modern systems
    if (S_ISDIR(stat.st_mode)) ...


nbriggs:
reverted master -- needed another change.

PR #422, opened 2022-05-24
Create missing directories when opening directory/file on {UNIX} device for output

nbriggs:
Prior to this change there was no mechanism within Medley to create a new directory
on a {UNIX} style device.  This change makes the {UNIX} directory creation happen in
the same manner that {DSK} directory creation is done.  Opening a file with access
OUTPUT, BOTH, or APPEND, which would create the file if it does not exist, will also
create any missing directories in the path.


nbriggs:
Fixes Interlisp/medley#779

rmkaplan (approved):
I tested, seems to solve the problem.  Thanks!

PR #423, opened 2022-05-27
fix some compiler warnings

nbriggs:
Just the smallest start on reducing the number of compiler warnings when compiling with -Wall -Weverything.  Getting the non-essential ones out of the way will make it easier to deal with any that are "real".


masinter (approved):
not sure why this PR languished. LGTM

PR #426, opened 2022-07-16
Prevent X11 error when resuming after call to RAID when -noscroll is in use

nbriggs:
The device_before_raid() call that disables X11 scrolling and bit-gravity
selection needs to avoid referring to the scrollbars and other ancillary
windows that have not been initialized when the "-noscroll" option was
given at startup.

Closes interlisp/medley#835


masinter (approved):
fixes problem, lgtm

PR #427, opened 2022-07-19
some cleanups in the arithmetic operations code

nbriggs:
Has no effect on the results, but cleans up some compiler warnings when warning levels are turned up.


masinter (approved):
we should add a step in the automatic build checks to run a test sysout.
but scanned the changes and lgtm

PR #432, opened 2022-07-24
quick_stack_check() should return a result indicating status

nbriggs:
When compiling with FNSTKCHECK enabled, quick_stack_check() is expected to return a result indicating whether the stack check found problems (1) or not (0).

Also corrects a missing include in xc.c which will cause errors when compiling with the FNSTKCHECK or FNTRACE.

### DIFF
--- a/inc/llstkdefs.h
+++ b/inc/llstkdefs.h
@@ -9,7 +9,7 @@ void flip_cursorbar(int n);
 void blt(register DLword *dest68k, register DLword *source68k, int nw);
 void stack_check(StackWord *start68k);
 void walk_stack(StackWord *start68k);
-void quick_stack_check(void);
+int quick_stack_check(void);
 void check_FX(FX *fx68k);
 void check_BF(Bframe *bf68k);
 int check_stack_rooms(FX *fx68k);
--- a/src/llstk.c
+++ b/src/llstk.c
@@ -829,7 +829,7 @@ void walk_stack(StackWord *start68k) {
                 for DEBUGING using FNSTKCHECK
 */
 /**************************************************************/
-void quick_stack_check(void) {
+int quick_stack_check(void) {
   StackWord *start68k;
   StackWord *scanptr68k;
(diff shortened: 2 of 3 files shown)


nbriggs:
@masinter -- could you sanity check the code that generates "WARN: CURRENTFX >= CurrentStackPTR??" -- when that check is enabled it fires constantly in call and return for lots of functions, e.g.

```
WARN: CURRENTFX >= CurrentStackPTR??
 
CURRENTFX=0x1237c,CurrentStackPTR=0x122b6
Calling a 0-arg FN:  IL:\EVENTKEYS().

WARN: CURRENTFX >= CurrentStackPTR??
 
CURRENTFX=0x1237c,CurrentStackPTR=0x122b6
In RETURN.
```
or
```
WARN: CURRENTFX >= CurrentStackPTR??
 
CURRENTFX=0x126b0,CurrentStackPTR=0x126a8
Calling a 2-arg FN:  IL:TIMEREXPIRED?(1669204640, IL:TICKS).

WARN: CURRENTFX >= CurrentStackPTR??
 
CURRENTFX=0x126b0,CurrentStackPTR=0x126a8
In RETURN.
```

masinter:
I don't remember much about the stack handling.
It would help to be able to read the documentation
(in Japanese, doesn't print to postscript)
image

nbriggs:
Do you have the (free) Google Translate app on your phone (either iOS or Android) -- in Camera mode, with translation from Japanese to English, it will do a pretty good job of translating.

masinter (approved):
These changes LGTM. I'm not sure about the curerentstack pointer question.
I want to see what happens when you have an APPLY.
It might also be the multiple-value-return problem? At least looking at the code for calls and returns that cause the warning.

nbriggs:
I suspect the frequent-firing check is not well-formed, but since it's turned off by default, I'll merge this and then we can look to fix that in the future.

PR #434, opened 2022-08-02
Update memory allocator function

nbriggs:
Update from using valloc() to posix_memalign() for larger memory allocations.

valloc() is deprecated, and not available on some systems (Android w/ bionic library) whereas posix_memalign() is available on all systems that we currently compile for.


waywardmonkeys:
LGTM

PR #436, opened 2022-08-09
Cleanup of includes and related changes based on include-what-you-use diagnostics

nbriggs:
Remove unused #define PERCENT_OF_SCREEN in MyWindow.h
Move structures for dir.c to dirdefs.h where they are used
Resolve S_CHAR vs S_CHARACTER in favor of S_CHARACTER and cleanup #defines
Fix  = vs == bug in FSDEBUG code in dir.c
Eliminate duplicate/unused constant definitions in gcr.c
Declare static internal function bytecmp in mkatom.c
Update many source and include files to include headers for what they use


inc/dirdefs.h
```diff
@@ -1,6 +1,65 @@
...
+                       * Link to a next entry.  Last entry is
+                       * indicated by NULL pointer.
+                       */
+} FINFO;
+
+typedef struct dfinfo {
+  FINFO *head; /* Head of the linked FINFO structures. */
+  FINFO *next; /* FINFO structure generated next time. */
+} DFINFO;
+
```
masinter, line 62: was this moved from somewhere else? the diffs are unclear.

nbriggs: Yes, they used to be defined in dir.c, but FINFO is used as type for a parameter if you're compiling with FSDEBUG, so the definitions need to be (and should have been, anyway) in dirdefs.h instead of dir.c.

masinter (approved):
I suppose this is like LOADCOMP of files used and no big EXPORTS.ALL for C code.

nbriggs:
Yeah, pretty much.  It wasn't well organized, and still isn't great, but at least most of the source (I couldn't fix them all) identifies what it depends on.  Some of the divisions into different files no longer make sense -- the compilers used not to handle "large" files well.  I think, for example, all the arith*.c could be combined, likewise for the array*.c.  This would also reduce the number of header files.   Much of the duplicated boilerplate would go away, so the total volume of stuff to read would be reduced, too.

PR #437, opened 2022-08-10
Combine six small array opcode implementation files into one

nbriggs:
Combines the array..array6 files and their associated header files
into new arrayops.c/arrayopsdefs.h files, and adjusts the CMakeLists.txt
and old-style makefile-tail to reflect these changes.


nbriggs:
Noticed an issue with the DOS makefile fragment.

masinter (approved):
it was odd to see a new file with an old copyright notice. I suppose we should revisit our copyright notice policy based on the advice we received.

nbriggs:
Since I was combining multiple old files, each of which had the same copyright notice, without making any substantive changes (I deleted some comments and deleted some duplicate includes and removed some "register" attributes) I erred on the side of leaving the original large copyright notice.  I did strip out the outdated file identification comments which also had copyright notices.

PR #439, opened 2022-08-12
Correct -Wformat-pedantic warnings

nbriggs:
Many of the warnings were newly visible after the DBPRINT macro
was modified, but the pedantic warnings also caught existing printf
family usage which was not strictly correct.

Mostly replacing 0x%x with %p for pointers, adding (void *) for
those and other pointers, %td for ptrdiff_t, but also changing
some function parameters from UNSIGNED (uintptr_t) to int
where that was the more appropriate type.


masinter (approved):
didn't try compiling with tracing going on (maybe a README on how to do that?) but been running with this & reviewed.

PR #440, opened 2022-08-16
Start to address warnings from -Weverything

nbriggs:
Before and after warning counts --

```
1337 [-Wcast-align]                	1337 [-Wcast-align]
1051 [-Wsign-conversion]           	 667 [-Wsign-conversion]
 519 [-Wstrict-prototypes]         	 514 [-Wstrict-prototypes]
 232 [-Wmissing-variable-declarations]	 218 [-Wmissing-variable-declarations]
 197 [-Wimplicit-int-conversion]   	 186 [-Wimplicit-int-conversion]
 147 [-Wshorten-64-to-32]          	 131 [-Wshorten-64-to-32]
  84 [-Wpadded]                    	  84 [-Wpadded]
  79 [-Wconditional-uninitialized] 	  79 [-Wconditional-uninitialized]
  52 [-Wunused-macros]             	  48 [-Wunused-macros]
  41 [-Wunused-parameter]          	  41 [-Wunused-parameter]
  36 [-Wsign-compare]              	  25 [-Wsign-compare]
  35 [-Wunreachable-code-break]    	  35 [-Wunreachable-code-break]
  15 [-Wmissing-prototypes]        	  14 [-Wmissing-prototypes]
   9 [-Wundef]                     	   7 [-Wundef]
   9 [-Wredundant-parens]          	   9 [-Wredundant-parens]
   9 [-Wcast-qual]                 	   9 [-Wcast-qual]
   7 [-Wtautological-value-range-compare]	
   7 [-Wshadow]                    	
   7 [-Wfloat-equal]               	   7 [-Wfloat-equal]
   5 [-Wmissing-noreturn]          	   5 [-Wmissing-noreturn]
   5 [-Wdouble-promotion]          	   5 [-Wdouble-promotion]
   4 [-Wswitch-enum]               	   4 [-Wswitch-enum]
   4 [-Wdocumentation-unknown-command]	   4 [-Wdocumentation-unknown-command]
   2 [-Wunused-variable]           	   1 [-Wunused-variable]
   2 [-Wcomma]                     	
   1 [-Wpointer-arith]             	   1 [-Wpointer-arith]
```

### DIFF
--- a/src/lowlev2.c
+++ b/src/lowlev2.c
@@ -35,7 +35,7 @@
 
 ***********************************************************/
 
-LispPTR N_OP_addbase(register int base, register int offset) {
+LispPTR N_OP_addbase(LispPTR base, LispPTR offset) {
   base = POINTERMASK & base;
   switch ((SEGMASK & offset)) {
     case S_POSITIVE: return (base + (offset & 0x0000FFFF));
@@ -65,7 +65,7 @@ LispPTR N_OP_addbase(register int base, register int offset) {
 
 ***********************************************************/
 
-LispPTR N_OP_getbasebyte(register LispPTR base_addr, register int byteoffset) {
+LispPTR N_OP_getbasebyte(LispPTR base_addr, LispPTR byteoffset) {
   switch ((SEGMASK & byteoffset)) {
     case S_POSITIVE: byteoffset = byteoffset & 0x0000FFFF; break;
     case S_NEGATIVE: byteoffset = byteoffset | 0xFFFF0000; break;
@@ -93,7 +93,7 @@ LispPTR N_OP_getbasebyte(register LispPTR base_addr, register int byteoffset) {
 
 ***********************************************************/
 
-LispPTR N_OP_putbasebyte(register LispPTR base_addr, register int byteoffset, register int tos) {
+LispPTR N_OP_putbasebyte(register LispPTR base_addr, LispPTR byteoffset, LispPTR tos) {
   if (((SEGMASK & tos) != S_POSITIVE) || ((unsigned short)tos >= 256)) ERROR_EXIT(tos);
   switch ((SEGMASK & byteoffset)) {
     case S_POSITIVE: byteoffset &= 0x0000FFFF; break;
--- a/src/lowlev2.c
+++ b/src/lowlev2.c
@@ -93,7 +93,7 @@ LispPTR N_OP_getbasebyte(LispPTR base_addr, LispPTR byteoffset) {
 
 ***********************************************************/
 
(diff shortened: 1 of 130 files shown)


masinter:
not sure how to test this except to run with it on as many systems as I can...
Get others to try it?

nbriggs:
It's less about trying it (I've done that on macOS x64, FreeBSD x86, Solaris SPARC in 32-bit mode) and more about reading the code changes for correctness.   Each commit is intended to be small, except for the case of something like "remove register attribute from all remaining files", but in that case that's the only thing that should have changed in the commit.  The github web diff should show these up pretty well.

src/lowlev2.c
```diff
@@ -93,7 +93,7 @@ LispPTR N_OP_getbasebyte(register LispPTR base_addr, register int byteoffset) {
 
 ***********************************************************/
 
-LispPTR N_OP_putbasebyte(register LispPTR base_addr, register int byteoffset, register int tos) {
+LispPTR N_OP_putbasebyte(LispPTR base_addr, LispPTR byteoffset, LispPTR tos) {
```
masinter, line 96: not sure I understand why they were int before and LispPTR now

src/shift.c
```diff
@@ -19,15 +19,15 @@
 /*
  * XXX: it feels as though something is not clean here, looks like the
  * "int a" arguments are really LispPTR types, though perhaps it doesn't
- * matter.  NBriggs, May 2017
+ * matter.  NBriggs, May 2017 -- Yes. Replaced. NBriggs, Aug 2022
  */
```
masinter, line 23: oh I see :)

masinter (approved):
i  should have taken this one commit at a time. There were some int -> LispPTR changes that I saw you noticed in 2017.
Anyway, LGTM

nbriggs:
Yes, one commit at a time, from first to last, should almost always be the right way to look at the changes in a PR.   I don't always succeed, but I try to make each commit be a small conceptual change that captures all the files that ought to change together (e.g., change in parameter types for declaration and implementation commit together, or removing the register attribute from all files.)
Sometimes this means I amend commits to add a changed file I missed, or undo an edit so that I can commit something before moving ahead again.

src/lowlev2.c
nbriggs: Probably because back in 2017 (or earlier) when I looked at the code I didn't connect up the fact that the other arguments being passed were NOT just numbers but were actually "LispPTR" types, pulled from the stack, from which an integer values was going to be extracted.  In the real original code all Lisp objects were just treated as "int", all the machines were 32-bit, and arithmetic and byte/bit operations were just assumed to do something that would work without regard to signedness... and the compilers didn't diagnose any of that as a potential problem.

PR #441, opened 2022-08-29
Run the 'mkvdate' script relative to CMAKE_CURRENT_SOURCE_DIR

kuriboshi:
This removes the limitation that the build directory has to be in the source tree and can instead be located anywhere.

Fixes issue Interlisp/medley#921.


waywardmonkeys:
LGTM

nbriggs (approved):
Yes.

PR #442, opened 2022-08-31
Use mmap() in place of posix_memalign() to allocate Lisp memory

nbriggs:
posix_memalign() does not guarantee initializing the allocated
  memory to zero, which Lisp expects, so the code must memset() the
  entire allocated region to zero.  The effect of this is (generally) to
  force the allocation of RAM to the process, which is wasteful, since
  we normally start with only 4% of 256 MB in use for a full.sysout.
  Allocating memory with mmap, using MAP_ANON, guarantees that the
  memory is already zeroed (effectively mapping /dev/zero with copy-on-write)
  so it is not necessary to touch it before use.  This keeps the
  pre-allocated RAM to a minimum.

### DIFF
--- a/src/ldsout.c
+++ b/src/ldsout.c
@@ -13,8 +13,9 @@
 
 #include <fcntl.h>        // for open, O_RDONLY
 #include <stdio.h>        // for perror, fprintf, printf, stderr, sprintf
-#include <stdlib.h>       // for exit, free, malloc, posix_memalign
+#include <stdlib.h>       // for exit, free, malloc
 #include <string.h>       // for memset
+#include <sys/mman.h>     // for mmap
 #include <sys/stat.h>     // for stat, fstat
 #include <unistd.h>       // for lseek, read, close, getpagesize
 #include "adr68k.h"       // for Addr68k_from_LADDR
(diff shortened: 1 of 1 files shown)


masinter (approved):
looks good, thanks

waywardmonkeys (approved):
I had two concerns about this, but neither seem valid / an issue.

1) Would this impact future WebAssembly support? No. WebAssembly can implement the mmap with `MAP_ANON` as a call to malloc.
2) Is there a difference between `MAP_ANON` and `MAP_ANONYMOUS`. Nope.

So, LGTM.

nbriggs:
@waywardmonkeys Thanks for the review.  I also dithered on MAP_ANON vs. MAP_ANONYMOUS.  Then I found that 8 years ago, on the OpenBSD mailing list:

> The Austin Group this morning accepted proposed wording to standardize
both MAP_ANON and MAP_ANONYMOUS, recognizing that neither was clearly
more popular than the other among applications, and that there's
precedent in POSIX for simply standardizing multiple spellings for a
feature when both are common and trivial for implementations to
provide (e.g., PAGE_SIZE and PAGESIZE).

So I stopped worrying about it.

PR #443, opened 2022-09-01
Replace Addr68k related macros with NativeAligned inline functions

nbriggs:
Addr68k_from_LADDR always produced a 2-byte aligned pointer which
  was frequently cast to a type requiring 4-byte alignment.
  This commit changes Addr68k_from_LADDR uses to one of two new
  inline procedures, NativeAligned2FromLAddr or NativeAligned4FromLAddr,
  which produce a result with the appropriate alignment for the context.
  This permits checking for cases where the Lisp address is not
  appropriately aligned for the usage context, and localizes compiler
  warnings to these two procedures.

  Similarly, the Addr68k_from_StkOffset macros are replaced by
  NativeAligned2FromStackOffset and NativeAligned4FromStackOffset.

  NativeAligned4FromLPage replaces Addr68k_from_LPAGE as page address
  will always be at least 4-byte aligned.
  
  LAddrFromNative, LPageFromNative, and StackOffsetFromNative complete
  the set, replacing LADDR_from_68k, LPAGE_from_68k, and
  StkOffset_from_68K (note K not k) respectively.

### DIFF
--- a/inc/adr68k.h
+++ b/inc/adr68k.h
@@ -59,10 +59,10 @@ static inline LispPTR *NativeAligned4FromLPage(LispPTR LPage)
   return (LispPTR *)(Lisp_world + (LPage << 8));
 }
 
-static inline DLword StackOffsetFromNative(DLword *SAddr)
+static inline DLword StackOffsetFromNative(void *SAddr)
 {
   /* Stack offsets are expressed as an offset in DLwords from the stack base */
-  ptrdiff_t hoffset = SAddr - Stackspace;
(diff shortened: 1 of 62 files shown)


nbriggs:
It would be great if people could try running on this branch and let me know if there are any issues with this that don't show up in the master branch.   If there are no issues, I'll plan to actually merge it in about 3 weeks.

nbriggs:
I hope that the cast-align branch has been working, since I've seen no evidence to the contrary, and will merge it shortly.

PR #445, opened 2022-10-03
Dodo Nethub support

devhawala:
Adds support for accessing XNS services provided by [Dodo Services - XNS with Java](https://github.com/devhawala/dodo), see the file `using-dodo-networking-with-maiko.md` for details.

This addition was tested on the platforms  `linux.x86_64-x` and `cygwin.x86_64-x`. Using Dodo Nethub networking is enabled for these 2 platforms by default in the corresponding makefiles.    
Using it for others platforms should be possible as no platform specific devices or functionalities are used,

An emulation for asynchronous events (timer, i/o) was added for the platform `cygwin.x86_64-x` to prevent the X-Windows UI from freezing (missing timer interrupts) and for networking (missing async i/o interrupts for ingoing packets).


waywardmonkeys (changes requested):
A couple of broader comments:

* I think this should include the cmake configuration for adding a toggle for enabling/disabling this.
* Should the new doc file go into a new directory called `docs` so that we can feel free to add more there and not clutter up the root directory of the repository?

Is this something that we want to enable for all Linux x86_64 and Cygwin builds? Does it work on macOS?

nbriggs:
@waywardmonkeys -- I've checked that it compiles and runs on macOS (after having to install Java so I could run the dodo code it talks to) and compiles (after the last update for missing include file) on FreeBSD.

I forgot that we had `-DMAIKO_ENABLE_ETHERNET` in `inc/version.h` which sets it automatically for Solaris/SunOS5 -- I think that's going to need to come out and we should use (command line) options for CMake to configure which networking is desired.  I might have chosen `-DMAIKO_ENABLE_ETHERNET` and `-DUSE_NETHUB` (cf. `-DUSE_DLPI` or `-DUSE_NIT`, or, still on my list of things to do: `-DUSE_BPF`), though since the NetHub doesn't require a root process to access a raw ethernet device we don't need to build the `ldeether` helper.

I think entropy would be reduced/ease of understanding increased by pulling the common (between NetHub and raw ethernet) support code into a separate file, and then separate files for NetHub vs the raw ethernet (possibly split the different raw ethernet accesses out, too -- they're pretty hard to read with line-by-line `#ifdef`s)

Opinions?

waywardmonkeys:
I agree about the options.

And having the options more configurable and accessible is (another) reason that the cmake build system should replace the legacy one. :)

I suspect I agree about separating out some files, but didn't look to see if that would be hard for any reason.

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+    return (NIL);
+  }
+  log_debug(("ether_send() - begin\n"));
+
+  u_char *source = (u_char *)NativeAligned2FromLAddr(args[1]);
+  int byteCount = 2 * (0xFFFF & args[0]); /* words to bytes */
+
+  log_debug(("   source = 0x%08X , bytecount: %d bytes\n", (unsigned int)source, byteCount));
+
+#ifdef BYTESWAP
```
nbriggs, line 626: I need to check this, but it feels wrong.  The Lisp code should have assembled the packet in network byte order.  If it turns out that swapping is necessary, I'd suggest copy to a temporary buffer, swap, send, and discard rather than re-swapping (or swap on the fly during the copy)

src/main.c
```diff
@@ -454,6 +466,80 @@ int main(int argc, char *argv[])
...
+    }
+#endif /* MAIKO_ENABLE_NETHUB */
+
+#if defined(MAIKO_EMULATE_TIMER_INTERRUPTS) || defined(MAIKO_EMULATE_ASYNC_INTERRUPTS)
+    else if (!strcmp(argv[i], "-intr-emu-insns")) {
+      if (argc > ++i) {
+        errno = 0;
+        tmpint = strtol(argv[i], (char **)NULL, 10);
+        if (errno == 0 && tmpint > 1000) {
+          timerAsyncEmulationInsnsCountdown = tmpint;
```
nbriggs, line 531: Not a big fan of emu-insns -- I don't see the value in identifying the counter as being in "emulated instructions".

using-dodo-networking-with-maiko.md
```diff
@@ -0,0 +1,86 @@
+# Using Dodo-networking with Maiko
```
nbriggs, line 1: I agree with @waywardmonkeys that it would be better to put this in a docs/ subdirectory.

nbriggs (commented):
See the comments attached through the code.
Needs to be tested on a big-endian system before merging (I'll do that).

src/main.c
```diff
@@ -285,6 +285,18 @@ const char *helpstring =
  -help        Print this message\n";
 #endif /* DOS */
 
+#ifdef MAIKO_ENABLE_NETHUB
```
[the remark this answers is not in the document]
devhawala: done

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+*/
+
+static int loglevel = 0;
+static int logcnt = 0;
+
+void setNethubLogLevel(int ll) {
+  loglevel = ll;
+  printf("nh :: leglevel now: %d\n", loglevel);
+}
+
```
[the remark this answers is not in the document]
devhawala: done (changes to config-parameters are now logged with `DBPRINT`)

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+#define log_all(line) { if (loglevel < 1) \
+	{ printf("nh[%d]: ", logcnt++); printf line; }}
+
+/*
+**  --- utilities ------------------------------------------------------------------
+*/
+
+#define LONG_PTR_MASK ((long)(~((long)3)))
+
+static void dblwordsSwap(u_char* basePtr, int forBytes) {
```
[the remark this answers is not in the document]
devhawala: done (`dblwordsSwap` now calls `word_swap_page` after adapting for alignment)

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+    wordsPtr[0] = b3;
+    wordsPtr[1] = b2;
+    wordsPtr[2] = b1;
+    wordsPtr[3] = b0;
+    wordsPtr += 4;
+    wordCount--;
+  }
+}
+
+static void asyncFd(int fd) {
```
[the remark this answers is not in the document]
devhawala: done

BUT: `int_io_open()` did not work at all, because of a typo in preprocessor directive (`#elseif` instead of `#elif`), also fixed that

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+  }
+
+  log_debug(("  connecting to nethub\n"));
+
+  struct sockaddr_in hubaddr;
+  hubaddr.sin_family = AF_INET;
+  hubaddr.sin_addr.s_addr = *((unsigned long *)host->h_addr);
+  hubaddr.sin_port = htons(nethubPort);
+
+  ether_fd = socket(AF_INET, SOCK_STREAM, 0);
```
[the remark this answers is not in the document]
devhawala: done

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+  memcpy(ether_buf, &rcvBuffer[2], copyLen);
+
+#endif
+
+  ether_bsize = 0;
+
+#if BYTESWAP
+log_debug(("  recvPacket() :: IOPage->dlethernet[2] = 0x%04X (before)\n", IOPage->dlethernet[2]));
+
+#if 1 /* byte-order in OPage->dlethernet[2] does not seem to matter ... boolean ? */
```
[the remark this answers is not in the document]
devhawala: done (removed unused code)

src/ether.c
devhawala: left it this way, swapping bytes is absolutely necessary (it took me a long time to understand how this is swapped, as the data pointer points to the 2nd 16-bit word of a 32-bit doubleword, so the first 2 packet bytes are in fact _before_ the pointer and the first 2 bytes pointed to are not part of the packet)

src/main.c
devhawala: done (the naming was meant to be read as timerAsyncEmulation-InsnsCountdown, not timerAsync-EmulationInsns-Countdown; the name is now `insnsCountdownForTimerAsyncEmulation`)

src/ether.c
devhawala: for copying the packet to a buffer and then swapping the buffer:

- that's what i did initially, the i saw that the _swap-use-swap_-pattern was used in the modules `inet.c`, `dsk.c`, `chardev.c` or `unixcomm.c`, so i switched to this pattern...
- and so i left it this way, also as the data is partly _before_ the original data pointer

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+  }
+
+  if (ether_bsize < 1) {
+    log_debug(("  recvPacket() :: no receive buffer present -> return 0\n"));
+    return 0;
+  }
+
+  if (nextPacket == NULL) {
+
+    /* try to get the next packet(s) form network */
```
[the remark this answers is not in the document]
devhawala: done (2nd commit due to comment hidden by github...)

src/ether.c
```diff
@@ -98,6 +108,603 @@ int ETHEREventCount = 0;
...
+#else
+  log_debug(("  recvPacket() :: copying %d bytes to 0x%016X\n", copyLen, (unsigned long)ether_buf));
+
+  memcpy(ether_buf, &rcvBuffer[2], copyLen);
+
+#endif
+
+  ether_bsize = 0;
+
+#if BYTESWAP
```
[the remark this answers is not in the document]
devhawala: done (2nd commit due to comment hidden by github...)

src/ether_nethub.c
```diff
@@ -0,0 +1,599 @@
...
+	{ printf("nh[%d]: ", logcnt++); printf line; }}
+
+/*
+**  --- utilities ------------------------------------------------------------------
+*/
+
+#define LONG_PTR_MASK ((long)(~((long)3)))
+
+static void dblwordsSwap(u_char* basePtr, int forBytes) {
+  u_char* wordsPtr = (u_char*)((long)basePtr & LONG_PTR_MASK);
```
nbriggs, line 125: See inc/lsptypes.h WORDPTR and the access macros defined for regular and BYTESWAP cases.

src/ether_nethub.c
```diff
@@ -0,0 +1,599 @@
...
+ **********************************************************************/
+LispPTR ether_send(LispPTR args[])
+{
+  if (nethubHost == NULL) {
+    return (NIL);
+  }
+  log_debug(("ether_send() - begin\n"));
+
+  u_char *source = (u_char *)NativeAligned2FromLAddr(args[1]);
+  int byteCount = 2 * (0xFFFF & args[0]); /* words to bytes */
```
nbriggs, line 522: I'll be looking to fix this in the original ether code as well -- this is NOT how one is supposed to access a Lisp numeric value -- I believe there's no guarantee that it's a SMALLP, the code may need to fetch a FIXP.
`#include "lisp2c.h"` and `n = LispInt2CInt(args[0]);`

devhawala: done (it's `lisp2cdefs.h` and `LispIntToCInt`)

src/ether_nethub.c
devhawala: no, `WORDPTR()` won't be helpful here:

- byte-swapping must start at a double-word boundary here and reverse the order of all 4 bytes in the double-words
- in the BYTESWAP-case `WORDPTR()` "simply" inverts the 2nd-to-last bit in the address, i.e. lets the address point to "the other" 16 bit word in a double-word
- this works only if the address points into the middle of a double-word: should the address point to the beginning of a double-word, all goes wrong
- i don't know if it is guaranteed that the packet buffer address will always start in the middle of a double-word
- and i still need the double-word address to compute the correct double-word count to byte-swap

So i left this procedure mostly unchanged and just renamed the variables to clarify that it's about double-words.

The only confusing thing left ist the name of the procedure `word_swap_page` which (despite its name) works on 32 bit chunks (reversing the bytes in double-words and not 16 bit words) and (worse) allows to start at the 2nd word (i.e. in the middle) of a double-word. (confusing in my opinion only, but tested with the mentioned byte-swapping outcomes)

nbriggs:
Does cygwin support `sendmsg` -- that is, you could avoid a packet copy on send (and likewise on recv, though I've not made that change in this example) by doing this:
```
diff --git a/src/ether_nethub.c b/src/ether_nethub.c
index c6e6d22..fb99191 100644
--- a/src/ether_nethub.c
+++ b/src/ether_nethub.c
@@ -327,14 +327,16 @@ static int recvPacket() {
 
 /* -1: failed/not connected; >= 0: packet bytes sent */
 static int sendPacket(u_char *source, int sourceLen) {
+  struct msghdr msg = {0};
+  struct iovec iov[2];
+  u_char nhlength[2];
+
   if (ether_fd < 0) {
     log_debug(("  sendPacket() :: not connected to hub !!!\n"));
     return -1;
   }
 
-  u_char sndBuffer[2050];
-  
-  if (sourceLen < 14 || sourceLen > (sizeof(sndBuffer) - 2)) {
+  if (sourceLen < 14) {
     log_debug(("  sendPacket() :: invalid packet length: %d !!!\n", sourceLen));
     return -1;
   }
@@ -344,18 +346,16 @@ static int sendPacket(u_char *source, int sourceLen) {
     return -1;
   }
 
-  sndBuffer[0] = (sourceLen >> 8) & 0x00FF;
-  sndBuffer[1] = sourceLen & 0x00FF;
-  int resLen = 0;
-  u_char *sPos = source;
-  u_char *dPos = &sndBuffer[2];
-  while(resLen < sourceLen) {
-    *dPos++ = *sPos++;
-    *dPos++ = *sPos++;
-    resLen += 2;
-  }
+  nhlength[0] = (sourceLen >> 8) & 0x00FF;
+  nhlength[1] = sourceLen & 0x00FF;
+  iov[0].iov_base = nhlength;
+  iov[0].iov_len = 2;
+  iov[1].iov_base = source;
+  iov[1].iov_len = sourceLen;
+  msg.msg_iov = iov;
+  msg.msg_iovlen = 2;
 
-  int result = send(ether_fd, sndBuffer, sourceLen + 2, 0);
+  int result = sendmsg(ether_fd, &msg, 0);
   return result;
 }
 
```
(as an aside: it would have been nice if the dodo packet protocol had used a network byte order length so it could have been declared a uint16_t and been accessed with htons/ntohs.  endian.h with htole16/le16toh are not available everywhere)

nbriggs:
[Edit of above:] -- Duh, it is documented as network byte order, big-endian.

nbriggs:
So revised suggested change for the code above is:
```
diff --git a/src/ether_nethub.c b/src/ether_nethub.c
index c6e6d22..fc766a6 100644
--- a/src/ether_nethub.c
+++ b/src/ether_nethub.c
@@ -327,14 +327,16 @@ static int recvPacket() {
 
 /* -1: failed/not connected; >= 0: packet bytes sent */
 static int sendPacket(u_char *source, int sourceLen) {
+  struct msghdr msg = {0};
+  struct iovec iov[2];
+  uint16_t nhlength;
+
   if (ether_fd < 0) {
     log_debug(("  sendPacket() :: not connected to hub !!!\n"));
     return -1;
   }
 
-  u_char sndBuffer[2050];
-  
-  if (sourceLen < 14 || sourceLen > (sizeof(sndBuffer) - 2)) {
+  if (sourceLen < 14) {
     log_debug(("  sendPacket() :: invalid packet length: %d !!!\n", sourceLen));
     return -1;
   }
@@ -344,18 +346,15 @@ static int sendPacket(u_char *source, int sourceLen) {
     return -1;
   }
 
-  sndBuffer[0] = (sourceLen >> 8) & 0x00FF;
-  sndBuffer[1] = sourceLen & 0x00FF;
-  int resLen = 0;
-  u_char *sPos = source;
-  u_char *dPos = &sndBuffer[2];
-  while(resLen < sourceLen) {
-    *dPos++ = *sPos++;
-    *dPos++ = *sPos++;
-    resLen += 2;
-  }
+  nhlength = htons(sourceLen);
+  iov[0].iov_base = &nhlength;
+  iov[0].iov_len = 2;
+  iov[1].iov_base = source;
+  iov[1].iov_len = sourceLen;
+  msg.msg_iov = iov;
+  msg.msg_iovlen = 2;
 
-  int result = send(ether_fd, sndBuffer, sourceLen + 2, 0);
+  int result = sendmsg(ether_fd, &msg, 0);
   return result;
 }
 
```

nbriggs:
This fixes a SIGBUS on a big-endian 32-bit system (also tested on 64-bit little-endian system).  Uses only POSIX required fields in hostent.

```
diff --git a/src/ether_nethub.c b/src/ether_nethub.c
index 250cd70..27478f3 100644
--- a/src/ether_nethub.c
+++ b/src/ether_nethub.c
@@ -164,7 +164,7 @@ void connectToHub() {
 
   struct sockaddr_in hubaddr;
   hubaddr.sin_family = AF_INET;
-  hubaddr.sin_addr.s_addr = *((unsigned long *)host->h_addr);
+  hubaddr.sin_addr.s_addr = *(in_addr_t *)host->h_addr_list[0];
   hubaddr.sin_port = htons(nethubPort);
 
   ether_fd = socket(AF_INET, SOCK_STREAM, 0);
```

nbriggs:
https://github.com/Interlisp/maiko/pull/446 fixes the async i/o setup for Solaris so that the NetHub code will actually work there.   I *think* Cygwin has <sys/file.h>, so there shouldn't be any issue there.

devhawala:
> This fixes a SIGBUS on a big-endian 32-bit system (also tested on 64-bit little-endian system). Uses only POSIX required fields in hostent.
> 
> ```
> diff --git a/src/ether_nethub.c b/src/ether_nethub.c
> index 250cd70..27478f3 100644
> --- a/src/ether_nethub.c
> +++ b/src/ether_nethub.c
> @@ -164,7 +164,7 @@ void connectToHub() {
>  
>    struct sockaddr_in hubaddr;
>    hubaddr.sin_family = AF_INET;
> -  hubaddr.sin_addr.s_addr = *((unsigned long *)host->h_addr);
> +  hubaddr.sin_addr.s_addr = *(in_addr_t *)host->h_addr_list[0];
>    hubaddr.sin_port = htons(nethubPort);
>  
>    ether_fd = socket(AF_INET, SOCK_STREAM, 0);
> ```

done

devhawala:
> #446 fixes the async i/o setup for Solaris so that the NetHub code will actually work there. I _think_ Cygwin has <sys/file.h>, so there shouldn't be any issue there.

yes, it does: i added this include to `timer.c`, it compiled and worked

src/ether.c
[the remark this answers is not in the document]
nbriggs: Now with additional changes in int_io_open etc.

src/ether.c
nbriggs: The pointer is subject to interpretation... it's not intended to be a pointer that can be used without manipulation by the `GETxxxx` macros to point it at the appropriate place for the endianness of the machine it's running on.
I'm OK merging it as is, but this is an area that we should look at very carefully once that's done.

PR #446, opened 2022-10-10
Correct async I/O setup to work for Solaris

nbriggs:
Solaris does not define O_ASYNC, but does define FASYNC and FNDELAY, which it expects to be used in place of O_ASYNC in an fcntl() to set up asynchronous I/O on a socket.

This fix also generates a compiler warning if neither the O_ASYNC nor FASYNC/FNDELAY methods are available.

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -25,6 +25,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/file.h>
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Did you test this at runtime and it works?

I thought the more correct way to do this in Solaris was `I_SETSIG`, but that is probably ancient memories and things have changed in the last many years.

PR #451, opened 2023-02-07
fix README.md: need libx11-dev, not x11dev

masinter:
as per axiomcas@gmail.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Building requires `clang`, `make`, X11 client libraries (`libx11-dev`). For exam
 
 ``` sh
 $ sudo apt update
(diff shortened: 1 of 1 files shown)


nbriggs:
The name of the package depends on the distribution you’re running.  Please don’t change this until we have checked what it is for the various distros.

nbriggs:
The name of the package depends on the distribution you’re running.  Please don’t change this until we have checked what it is for the various distros.

nbriggs:
I'm OK with this change -- all the distros that use apt now seem to use libx11-dev for the client-side X11 headers.

PR #454, opened 2023-02-22
Adding a new workflow to build a docker image with the tools to compile maiko

fghalasz:
Need this added to master before I can test on my branch since workflows don't show up in actions unless they are in master.


fghalasz:
Ooops.  Looks like this PR pulled in too many commits.  Need to straighten out and resubmit.

PR #455, opened 2023-02-22
Add a workflow to build a Docker image with all tools to build Maiko for Linux

fghalasz:
Adding a new workflow (github action) to build a multi-platform Docker image with all the tools to build Linux Maiko.  This is part of making Maiko Release build workflow much more efficient.

This just adds first-pass at said workflow.  But need to integrate this into Master now, since without the workflow in master, it will not show up in github actions for testing on the development branch.

WIll have no effect on anything unless you execute this workflow and then - if successful -- it will only create a docker image and store it in ghcr.io.

### DIFF
--- a/.github/workflows/buildBuilderImage.yml
+++ b/.github/workflows/buildBuilderImage.yml
@@ -1,0 +1,93 @@
+#*******************************************************************************
+# buildBuilderImage.yml
+#
+# Workflow to build a multiarch Docker image that includes all of the tools for
+# building maiko for linux.  Intended to be used (exclusively?) by the github
+# actions that build maiko releases - e.g., buildRelease.yml.
+#
+# The purpose is to make the maiko github actions quicker and less resource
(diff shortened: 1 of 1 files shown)


nbriggs:
BTW - I've discovered that CMake will build the universal binaries (in theory on either architecture) just by filling in one define in the CMakeFiles.txt that is already there.

We were considering getting rid of the "makeright" script and settling on using CMake as the build tool.  It also makes it easier to select (either using the CMake-gui, or through defines) the various configuration options, such as what kind of network code to use, whether to compile for X11 or SDL display.

nbriggs (approved):
LGTM.

fghalasz:
Re: cmake:  I can easily switch over the release builds to cmake.  Is it ready to go?

I don't know cmake at all, but looking thru the cmake stuff, I didn't see any support for compiling ldeinit.  What is the cmake equivalent of makeright init ?

PR #456, opened 2023-02-22
Maiko code changes needed to compile on Alpine Linux with clang. 

fghalasz:
Switching to using Alpine Linux to compile maiko for Linux in the release build workflows.  Much more efficient than the bloated Ubuntu.

Needed to make small changes to two files to get to compile in Alpine:
1.  In inc/dirdefs.h:  MAXNAMLEN is BSD not Posix.  Posix is NAME_MAX.  So need to set MAXNAMLEN to NAME_MAX if MAXNAMLEN not defined.

2.  In src/ether_nethub.c:  need to add includes for <sys/select.h> and <sys/types.h> to get u_char, fd_set, and FD_SET defined.

### DIFF
--- a/inc/dirdefs.h
+++ b/inc/dirdefs.h
@@ -1,17 +1,9 @@
 #ifndef DIRDEFS_H
 #define DIRDEFS_H 1
 
-#include <dirent.h>         // for MAXNAMLEN
-
-/* Needed for Alpine Linux */
-#ifndef MAXNAMLEN
-#include <limits.h>
-#define MAXNAMLEN NAME_MAX
-#endif
-/* End Needed for Alpine Linux */
-
 #include <sys/types.h>      // for u_short, ino_t
 #include "lispemul.h"       // for LispPTR
+#include "locfile.h"        // for MAXNAMLEN
 /*
  * FINFO and FPROP are used to store the information of the enumerated files
  * and directories.  They are arranged in a form of linked list.  Each list is
--- a/inc/dirdefs.h
+++ b/inc/dirdefs.h
@@ -1,6 +1,15 @@
 #ifndef DIRDEFS_H
 #define DIRDEFS_H 1
+
 #include <dirent.h>         // for MAXNAMLEN
+
+/* Needed for Alpine Linux */
+#ifndef MAXNAMLEN
+#include <limits.h>
+#define MAXNAMLEN NAME_MAX
+#endif
+/* End Needed for Alpine Linux */
+
 #include <sys/types.h>      // for u_short, ino_t
 #include "lispemul.h"       // for LispPTR
 /*
--- a/src/ether_nethub.c
+++ b/src/ether_nethub.c
@@ -38,6 +38,12 @@
 #include "timerdefs.h"
 #include "lisp2cdefs.h"
 
+/* Needed for Alpine Linux */
+#include <sys/select.h>
+#include <sys/types.h>
+/* End Needed for Alpine Linux */
+
+
 /*
 **  --- ether implementation common data -------------------------------------------
 */
--- a/src/ether_nethub.c
+++ b/src/ether_nethub.c
@@ -14,8 +14,10 @@
 #include <string.h>
 #include <unistd.h>
 #include <sys/socket.h>
+#include <sys/select.h>
+#include <sys/types.h>
 #include <arpa/inet.h>
-#include <sys/poll.h>
+#include <poll.h>
 #include <signal.h>
 #include <fcntl.h>
 #include <errno.h>
@@ -37,12 +39,6 @@
 #include "dbprint.h"
 #include "timerdefs.h"
 #include "lisp2cdefs.h"
-
-/* Needed for Alpine Linux */
-#include <sys/select.h>
-#include <sys/types.h>
-/* End Needed for Alpine Linux */
-
 
 /*
 **  --- ether implementation common data -------------------------------------------
--- a/inc/locfile.h
+++ b/inc/locfile.h
@@ -9,6 +9,7 @@
 /*									*/
 /************************************************************************/
 #include <errno.h>
+#include <limits.h>   /* for NAME_MAX */
 #include "lispemul.h" /* for DLword */
 
 #define	FDEV_PAGE_SIZE		512	/* 1 page == 512 byte */


src/ether_nethub.c
```diff
@@ -38,6 +38,12 @@
 #include "timerdefs.h"
 #include "lisp2cdefs.h"
 
+/* Needed for Alpine Linux */
+#include <sys/select.h>
```
nbriggs, line 42: Could you say which symbols these are needed for?

fghalasz: <sys/select.h> has defs for fd_set and FD_SET()
<sys/types.h> has typedef for u_char

I think its Posix standard for fd_set and FD_SET to be defined in sys/select.h, but a lot of platforms also define then elsewhere (like types.h).  Not so for Alpine.

And u_char is also not Posix but a BSD thing that most platforms define in various header files.  But on Alpine, it requires loading the sys/types.h header file.

I think all this stems from Alpine (and several other distros) using the musl package for libc instead of the glibc package, which is found on most of the "bigger" Linux (and apparently Mac) platforms.

nbriggs: OK.  I'm just installing an Alpine Linux VM so I can test maiko things.

I'd like to move the code as close to POSIX as possible, while still compiling on things like Android, FreeBSD, other *BSD, Solaris 11/SPARC, MacOS/PPC, etc.

waywardmonkeys:
I am on the road and on my phone, so please forgive formatting and the like.

1) I think a separate PR to just use `NAME_MAX` would be better.
2) I think switching away from `u_char` your `uint8_t` or `unsigned char` or whatever would better.
3) macOS doesn’t use GNU libc.
4) the `sys/select.h` header is correct everywhere, I think. Shouldn’t need a comment about Alpine. It would also be true on other platforms.

fghalasz:
@waywardmonkeys:  Agree with your items 1 & 2, but they were much bigger changes that I wanted to do as only a passer-by w.r.t the maiko code.  But note that from what I could see the whole MAXNAMLEN situation in maiko is a little bit confused since it sometimes appears to be used as a synonym of NAME_MAX and sometimes appears to be used in place of MAX_PATH.  So some more extended cleanup is needed.

On item 4:  It looks like on (some) other platforms (e.g., Ubuntu and I guess MacOS), \<sys/types.h>  \#includes \<sys/select.h>.  That's why we haven't run into this issue before now.

waywardmonkeys:
If there is no rush, I can do some of these stuff ver the next 2-3 days.

nbriggs:
Let's have a discussion about exactly what to do before we do it.   NAME_MAX and PATH_MAX are two different things -- NAME_MAX is (should be) the maximum length of the name of a file (or an isolated directory?), not considering the directories in which it may be contained, while PATH_MAX is the total length, from root to the file.

The maiko file management code may (I haven't looked before writing this) depend on both, and *should* be clear about which one is in play in any given circumstance.

I agree that we should be using NAME_MAX and PATH_MAX and we'll need to compensate for systems that don't have that predefined (and have MAXNAMLEN or MAXNAMELEN, and MAXPATHLEN).

It's still the case that "char" may be either signed or unsigned, so it's definitely the case that where the code means to treat a value as a small integer it should use int8_t or uint8_t (as appropriate) rather than "char" or "unsigned char" or "u_char".

So, I think my position is:

1. Yes, 2. Yes, 3. I agree, 4. Yes.

Part of the issue on Alpine is which packages you install for the header files (and compilers and ancillary tools).   I started with a plain Alpine Linux image, and did:
```
   apk add git libx11-dev make binutils
```
now I'm looking at `musl-dev`, and `gcc` (and you need `gcc` if you want to use `clang` because it depends on the libgcc[_s] and crt*.o from gcc).

There are also compatability packages for BSDisms (which we should probably avoid if we can clean up the other things).    I haven't finished my exploration of this yet (and it will take a day or two more, elapsed time, I think)

In making (these and other...) changes, please note that we've had people compiling on:  macOS (little- (x86, x86_64, aarch64) and big-endian (ppc32) hardware), FreeBSD (i386 and x86_64), Solaris 11 (SPARC), Android (using termux, I believe), and various Linux distros including Ubuntu specifically, other random Debian variants, and now Alpine Linux.  While I can compile on some of these, I don't have access to all of them, so it's a bit difficult to validate the changes.

nbriggs:
P.S., on item 4 -- on Solaris I run include-what-you-use against (many of) the sources and includes, and where it doesn't get completely confused and produce bogus results I use it as a guide for which include files should be there.

While I'm trying to complete the include-what-you-use pattern, there are certainly circumstances where the code (still) relies on incidental nested includes.

nbriggs:
MAXNAMLEN is defined in sys/param.h on Alpine Linux, and is available in the musl-dev package.

fghalasz:
Is <sys/param.h> guaranteed to exist on all other platforms (that we care about)?  And is it benign to `#include <sys/param.h>` in inc/dirdefs.h even on platforms where MAXNAMLEN is defined elsewhere?

nbriggs:
Solaris doesn't define NAME_MAX anywhere.  Sigh.

I also noticed that maiko's inc/locfile.h checks
```
#if !defined(MAXNAMLEN)
#define MAXNAMLEN NAME_MAX
#endif
```
but does not include <limits.h>, which it presumably should since it may be depending on NAME_MAX.

So, I would suggest before trying to do a shift towards NAME_MAX, without breaking Solaris builds, the right thing might be:

```
diff --git a/inc/dirdefs.h b/inc/dirdefs.h
index 260c23e..6f72fd3 100644
--- a/inc/dirdefs.h
+++ b/inc/dirdefs.h
@@ -1,6 +1,6 @@
 #ifndef DIRDEFS_H
 #define DIRDEFS_H 1
-#include <dirent.h>         // for MAXNAMLEN
+#include "locfile.h"         // for MAXNAMLEN
 #include <sys/types.h>      // for u_short, ino_t
 #include "lispemul.h"       // for LispPTR
 /*
diff --git a/inc/locfile.h b/inc/locfile.h
index b0fda81..a97b72f 100644
--- a/inc/locfile.h
+++ b/inc/locfile.h
@@ -9,6 +9,7 @@
 /*									*/
 /************************************************************************/
 #include <errno.h>
+#include <limits.h>  /* for NAME_MAX */
 #include "lispemul.h" /* for DLword */
 
 #define	FDEV_PAGE_SIZE		512	/* 1 page == 512 byte */
```

and to accept your modification to ether_nethub.c but without the Alpine Linux comments because it *should* have had those includes anyway (as so many other source files already do).

nbriggs:
For ether_nethub.c -- Compiling on Alpine Linux also points out that `#include <sys/poll.h>` should be just `#include <poll.h>` (which is correct for other systems as well).

nbriggs:
On Solaris, this is what some code does that depends on NAME_MAX:
```
/* NAME_MAX is used to compute the allocation size for a struct dirent
 * passed to readdir() / readdir_r(). However on some systems it is not
 * defined, as it is technically a filesystem dependent property that
 * we should retrieve through pathconf(). In any case, we do not need a
 * precise value but only an upper limit.
 */
#ifndef NAME_MAX
# ifdef MAXNAMELEN
   /* Solaris has no NAME_MAX but defines MAXNAMELEN */
#  define NAME_MAX MAXNAMELEN
# elif defined(PATH_MAX)
   /* PATH_MAX (maximum length of a full path name) is a safe fall back */
#  define NAME_MAX PATH_MAX
# elif defined(FILENAME_MAX)
   /* Similarly FILENAME_MAX can provide a safe fall back */
#  define NAME_MAX FILENAME_MAX
# else
   /* Hardcode a reasonably large value as a last chance fallback */
#  define NAME_MAX 1024
# endif
#endif
```

nbriggs:
... also odd, there's both `MAXNAMLEN` and `MAXNAMELEN`.    Both are defined in different places on Solaris, and have different values.  I see other bits of code that in the absence of `NAME_MAX` use `PATH_MAX/2`.  Let's hear it for standards... there are so many to choose from.

nbriggs:
I've updated the branch to my preferred solution for this, and fixed the <sys/poll.h> -> <poll.h> at the same time.

I've tested compilation on my Alpine Linux VM, and it compiles cleanly.  After a second set of eyes on it, I'm happy to merge it.

We can address the overall use of NAME_MAX vs MAXNAMLEN in a different commit.

inc/dirdefs.h
```diff
@@ -1,17 +1,9 @@
...
-/* Needed for Alpine Linux */
-#ifndef MAXNAMLEN
-#include <limits.h>
-#define MAXNAMLEN NAME_MAX
-#endif
-/* End Needed for Alpine Linux */
-
 #include <sys/types.h>      // for u_short, ino_t
 #include "lispemul.h"       // for LispPTR
+#include "locfile.h"        // for MAXNAMLEN
```
waywardmonkeys, line 6: Why not just include `<limits.h>` here as well? I am on my phone so don’t have full context readily available…

nbriggs: I did that because `<limits.h>` will get you `NAME_MAX`, but not `MAXNAMLEN` and `"locfile.h"` has the `ifndef MAXNAMLEN` logic to define it as `NAME_MAX`, which I wanted to keep in just one place until we do the wholesale replacement of MAXNAMLEN with NAME_MAX (plus the logic for when NAME_MAX is not defined, such as on Solaris)

waywardmonkeys: Makes sense.

PR #458, opened 2023-02-26
Remove unused `struct lbits` / `LBITS`.

waywardmonkeys:


nbriggs (approved):
LGTM -- I see no evidence of it having been used even back in 1988.

PR #461, opened 2023-03-12
Speed up initial sysout loading by avoiding unnecessary lseek() calls.

nbriggs:
On slower filesystems/disks, an `lseek()` operation can consume noticeable time. During initial sysout loading avoid an unnecessary `lseek()` before each `read()` by tracking if the place it will `lseek()` to is the same place that the previous `read()` left us positioned.

### DIFF
--- a/src/ldsout.c
+++ b/src/ldsout.c
@@ -17,6 +17,7 @@
 #include <string.h>       // for memset
 #include <sys/mman.h>     // for mmap
 #include <sys/stat.h>     // for stat, fstat
+#include <sys/types.h>    // for off_t
 #include <unistd.h>       // for lseek, read, close, getpagesize
 #include "adr68k.h"       // for NativeAligned2FromLAddr
 #ifdef BYTESWAP
@@ -74,6 +75,7 @@ unsigned sysout_loader(const char *sysout_file_name, unsigned sys_size) {
   struct stat stat_buf; /* file stat buf */
 
   char errmsg[255];
+  off_t cfp = -1;  /* tracks current file position in sysout, or -1 */
 
   /* Checks for specifying the process size (phase I) */
   /* If sys_size == 0 figure out the proper size later */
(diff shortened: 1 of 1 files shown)


src/ldsout.c
```diff
@@ -74,7 +74,8 @@ unsigned sysout_loader(const char *sysout_file_name, unsigned sys_size) {
   struct stat stat_buf; /* file stat buf */
 
   char errmsg[255];
-
+  int cfp = -1;
```
waywardmonkeys, line 77: Should this be `int` or `off_t`?

waywardmonkeys:
It would be nice if the comment in this PR was in the commit comment ... and perhaps a comment in the code around `cfp` or something ...

src/ldsout.c
nbriggs: Yes, it probably should be `off_t` -- I knew it was at most a 256 MB file (that limit is baked into the Lisp memory structure because of bitfield sizes that will not change).

I'll fix that.

nbriggs:
Agreed, thanks - I had very limited time when I pushed this for review this AM.  I'll fix the commit comment and comment the code.

masinter:
when recompiling the system a "clean" compile where you reload the sysout each time -- would be after
or TEdit print from a server. A logout(T) + reload would be great. If optionally you could change the screen size that would be even better.

nbriggs:
@masinter - I'm not quite sure what you're suggesting.  This speeds up the reading of a sysout back into Lisp's virtual memory whenever you start lde/ldex.  For most people that's _almost_ instantaneous already, but on an old machine I was running each `lseek()` took about twice as long as the `read()` and did nothing at all since the read-head was already at the right place. It was on slow spinning rust storage. Easy fix, so I'm speeding it up without changing any functionality at all at the moment.

waywardmonkeys:
@nbriggs Sounds great! Happy to review further after that.

nbriggs:
I think this could be re-reviewed now.

PR #464, opened 2023-03-29
Support for netbsd

mldulaney:
Adds support for netbsd

### DIFF
--- a/bin/makefile-netbsd.x86_64-x
+++ b/bin/makefile-netbsd.x86_64-x
@@ -1,0 +1,28 @@
+# Options for NetBSD, Intel x86_64 and X-Window
+
+CC = clang -m64 $(CLANG_CFLAGS)
+
+LD_LIBRARY_PATH = /usr/lib
+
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Does the cmake build system work here without changes?

nbriggs (changes requested):
If you add the “init” version of the makefile fragment as well then it will enable doing loadups on netbsd also.  See the FreeBSD equivalent…

mldulaney:
> Does the cmake build system work here without changes?

It did not seem to make all binaries



> If you add the “init” version of the makefile fragment as well then it will enable doing loadups on netbsd also. See the FreeBSD equivalent…

Got that bit working, too, and I'll update my PR shortly.

nbriggs:
@mldulaney - the cmake default is to make only the binaries you need to run an existing sysout, so `lde` and `ldex`.  Given that we're not checking in sysout files these days it would make sense to have the CMake targets include `ldeinit` as well.

mldulaney:
using cmake worked.  closing.

PR #466, opened 2023-05-05
Remove "katana" from maiko files

nbriggs:
The "katana" project name predated the "maiko" project name for the Lisp virtual machine implementation.  There were only a few references left in the virtual machine code which are here replaced by equivalent constants/comments using Maiko. There are no functional or user-visible differences.


nbriggs:
Closes https://github.com/Interlisp/medley/issues/122

PR #469, opened 2023-06-05
Update github actions to support cross-compiling maiko on Darwin; update darwin makefiles to support same, ...

fghalasz:
Also update github actions with latest action versions and set-output schemas.


fghalasz:
Closing this pull request; will rebase on latest master and update Darwin-SDL builds as well.

PR #470, opened 2023-06-06
Workflow updates for Darwin: add SDL to release builds; enable cross-compile in makefiles; add cross-compile & universal executables to release builds ...

fghalasz:
Also update workflows to latest action versions and update handling of GITHIUB_ENV to get rid of deprecated set-output.

This is a resubmit of closed PR#469.  Changes from that PR include rebased branch onto latest master; added support for SDL in release builds for Darwin.

### DIFF
--- a/.github/workflows/buildReleaseInclDocker.yml
+++ b/.github/workflows/buildReleaseInclDocker.yml
@@ -309,7 +309,7 @@ jobs:
         run: |
           gh release download release-2.26.5 --repo libsdl-org/SDL --pattern SDL2-2.26.5.dmg
           hdiutil attach SDL2-2.26.5.dmg
-          cp -r /Volumes/SDL2/SDL2.framework /Library/Frameworks
+          ditto /Volumes/SDL2/SDL2.framework /Library/Frameworks/SDL2.framework
           hdiutil detach /Volumes/SDL2/
 
       # Build maiko   
--- a/.github/workflows/buildReleaseInclDocker.yml
+++ b/.github/workflows/buildReleaseInclDocker.yml
@@ -139,9 +139,9 @@ jobs:
   linux:
   
     needs: [inputs, sentry]
-    if: |
-      needs.sentry.outputs.release_not_built == 'true'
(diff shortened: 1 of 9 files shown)


nbriggs (approved):
This looks OK to me.  I don't know enough about the workflow to make any significant comments on it.   The effect of using "makeright x" then "makeright sdl" that you'll end up with an "lde" that can only start an "ldex".  Currently, the only way to get an "lde" that can start either one (depending on the "-display" option) is to build using CMake.

masinter:
> This looks OK to me. I don't know enough about the workflow to make any significant comments on it. The effect of using "makeright x" then "makeright sdl" that you'll end up with an "lde" that can only start an "ldex". Currently, the only way to get an "lde" that can start either one (depending on the "-display" option) is to build using CMake.

should this be a new (low priority) issue?

nbriggs:
Fixing the "makeright" so you can do both?  It would add another "display" target.   Not sure that it's worth it when you can compile the whole thing, making universal binaries, with both X11 and SDL2 by doing:
```
cd maiko
mkdir build
cmake .. -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12 -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64" -DMAIKO_DISPLAY_SDL=ON -DMAIKO_DISPLAY_X11=ON
```
which will leave you with lde, ldex, ldesdl (I don't have it doing ldeinit yet) in the build directory.

fghalasz:
Ok.  Switched over build of lde, ldex and ldesdl to cmake.,   Left ldeinit to be made by makeright init.

Also changed Xquartz to be loaded by installer of XQuartz-2.8.5.pkg rather than via homebrew.  And removed all the xlibs that are preinstalled on MacOS runners cause they were interfering with cross-compile of aarch64 on x86_64 runners.

nbriggs:
Nice.  I've got a CMakeLists.txt that defines how to also compile ldeinit when you have X11 enabled (I can't do an SDL based ldeinit).  I'll probably put the PR in on Friday.

PR #472, opened 2023-06-08
First step in renaming workflow buildReleaseInclDocker to just buildRelease

fghalasz:
.  This commit allows us to test changes to buildRelease workflow on a branch before making the name change permanent.

This renaming is because we are no longer making docker images for maiko releases.


fghalasz:
Trivial change in workflow code.  No need for review.

PR #473, opened 2023-06-08
Github workflow: increase efficiency of Linux builds (slightly)

fghalasz:
Change how buildRelease does Linux builds to make them slightly more efficient.  Does this by using a docker builder image that has all the tools preloaded rather than loading them each time buildRelease runs.


nbriggs (approved):
I don't see anything obviously wrong.  It would be helpful to have an overview of what this whole build process is attempting to accomplish - is it possible to have a README(.md) in the .github/workflows directory?  That's not a gate for merging this, but would be helpful for those of us that don't deal in github workflows and Docker (and qemu) every day.

PR #477, opened 2023-07-24
struct buf is missing definition for byte-swapped bigvm case

nbriggs:
While there was a definition for the pre-bigvm case, with 24-bit pointers, for byteswapped (little-endian) systems there was no structure definition for the bigvm case, with 28-bit pointers.

### DIFF
--- a/src/gcfinal.c
+++ b/src/gcfinal.c
@@ -97,6 +97,18 @@ struct buf {
 };
 #endif /* BIGVM */
 #else
+#ifdef BIGVM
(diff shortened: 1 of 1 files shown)


masinter (approved):
If the declaration was missing, how did it compile?
The changes look good as far as I can tell.

nbriggs:
For little-endian machines it handled the BIGVM case as the not-BIGVM case so you got a definition, it was just wrong for the situation.

PR #479, opened 2023-07-31
Update machinetype and osversion scripts to invoke config.guess from same directory

nbriggs:
If machinetype or osversion are executed from other than the current working directory they will fail to execute the config.guess script because it was invoked with "./config.guess". This update attempts to determine the directory where machinetype or osversion were found and uses the same path to invoke config.guess.

### DIFF
--- a/bin/machinetype
+++ b/bin/machinetype
@@ -9,7 +9,8 @@
 #									#
 #########################################################################
 
-os=${LDEARCH:-`./config.guess`}
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+os=${LDEARCH:-`$SCRIPTPATH/config.guess`}
 # o/s switch block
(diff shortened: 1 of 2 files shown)


masinter (approved):
LGTM
This whole mechanism is pretty hackish. Does cmake help?

waywardmonkeys:
> LGTM This whole mechanism is pretty hackish. Does cmake help?

Yes, it doesn't need these scripts.

nbriggs:
Cmake avoids *that* hackishness, substituting for it things you write in the CMakeLists.txt.

I suspect a couple of reasons for the existing makefile hackishness - (1) the compiler name and options really were significantly different per architecture and OS combination, and (2) all the executables got collected together into one set of release media (or fileserver) and then the appropriate set was pulled out and installed, so a directory named for each combination made sense, and those needed to be standardized, thus the two scripts that generated the expected names.

PR #481, opened 2023-07-31
Avoid setting Lisp variables whose locations are unknown when in INIT mode

nbriggs:
The X11 event handling will attempt to update the last user action time and the mouse/cursor position Lisp variables, but when the code is running in INIT mode the native addresses for those variables are not initialized and so we end up with a segmentation fault. Just don't attempt those updates when compiled with -DINIT

### DIFF
--- a/src/vmemsave.c
+++ b/src/vmemsave.c
@@ -531,5 +531,5 @@ void lisp_finish(void) {
 #ifdef DOS
   exit_host_filesystem();
 #endif /* DOS */
-  exit(1);
+  exit(0);
 }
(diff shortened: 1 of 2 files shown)


fghalasz:
LGTM.  But it didn't fix the crash I was having in loadup-mid-from-init.sh.   Have run multiple loadups-all.sh with this change and they have all succeeded.

In terms of the crash in loadup-mid-from-init.sh:  turns out its been happening for quite some time.  But the crash happens during LOGOUT after the needed vmem file has been created/saved - so its a no-op from loadups-all pov.  I just changed my modified loadup-mid-from-init.sh script to exit success despite the crash as the previous script had done.

nbriggs:
Odd… it fixed the crash I saw at startup, and I’m not seeing a crash when it exits.  Can you run it under a debugger and give me a stack trace when it crashes?

nbriggs:
Ah, is it because exiting lisp is always with exit(1), which is EXIT_FAILURE ?  IMO that’s a bad choice for the “normal” exit status - how about I change it to 0 (EXIT_SUCCESS)?

PR #482, opened 2023-08-08
More Lisp variables to avoid in INIT mode - turn off X11 input handling

nbriggs:
For the lde handling the INIT processing, no keyboard input is necessary and in fact any keyboard or mouse button presses will cause the code to take a segmentation fault attempting to store keystrokes into Lisp variables that the emulator doesn't have a good reference for.  So, in the X11 event processing when in INIT mode, ignore key presses and mouse movement.


fghalasz (approved):
LGTM.  

Can no longer cause failures in loadup-mid-from-init.sh by keypresses.

PR #485, opened 2023-11-26
Haiku OS port

Sylvain78:
A port to [Haiku](www.haiku-os.org) on x86_64 with the SDL backend.


src/initsout.c
```diff
@@ -140,9 +140,9 @@ void init_ifpage(unsigned sysout_size) {
 #endif /* BIGVM */
 
   /* unfortunately, Lisp only looks at a 16 bit serial number */
-#ifndef DOS
+#if !defined(DOS) && !defined(MAIKO_OS_HAIKU)
   InterfacePage->serialnumber = 0xffff & gethostid();
```
nbriggs, line 144: I'm surprised Haiku doesn't have a `gethostid()` since the [libroot documentation](https://www.haiku-os.org/docs/api/libroot.html) says it implements [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/) calls which include `gethostid()`

src/ldeboot.c
```diff
@@ -200,7 +200,12 @@ int main(int argc, char *argv[]) {
     argv[0] = filetorun;
   } else {
     /* copy up to and including the final "/" in the path */
+#ifndef MAIKO_OS_HAIKU
```
nbriggs, line 203: `strncpy(dst, src, ...)` is documented as returning `dst`, while `stpncpy(dst, src, ...)` is documented as returning a pointer to the terminating NUL (if there is one) of `dst`.  They are NOT interchangeable.

src/initsout.c
lassik: @nbriggs https://dev.haiku-os.org/ticket/10841

nbriggs: We should also open an issue against the libroot documentation to suggest that it include a list of unimplemented features rather than just referring to the POSIX spec and then not implementing some of it.  That would have shortened the journey to this answer.

The gethostid() call used to (on Sun hardware) return the hardware serial number.  The underlying implementation changed, but fairly early on it moved from its own system call to be a compatibility function loosely based on the (SunOS/Solaris) sysinfo() call, SI_HW_SERIAL, result
>       SI_HW_SERIAL

           Copy  into  the array pointed to by buf a string which is the ASCII
           representation of the hardware-specific serial number of the physi-
           cal machine on which the function is executed. This might be imple-
           mented in Read-Only  Memory,  using  software  constants  set  when
           building the operating system, or by other means, and might contain
           non-numeric characters. If the function is executed within  a  non-
           global  zone that emulates a host identifier, then the ASCII repre-
           sentation of the zone's host identifier is copied  into  the  array
           pointed  to  by  buf. It is anticipated that manufacturers will not
           issue the same "serial number" to more than one  physical  machine.
           The  pair of strings returned by SI_HW_PROVIDER and SI_HW_SERIAL is
           not guaranteed to be unique across all  vendor's  SVR4  implementa-
           tions and could change over the lifetime of a given system.

I'm not sure what glibc implementors were thinking when they based it on the IP address - that would never have been a reasonable thing to do

nbriggs: A point to note in all this, though, is that the only place the "hostid" now shows up is in the common lisp environment code where it effectively tries to do a gethostid() and if that fails it tries for HOSTID (as a string) in the Unix environment.  Maiko used to be licensed software, based on the hostid, but that is long gone.  The right thing would be to use the system-uuid, but there's no standardized (and unprivileged) way of determining that on the variety of systems that we run on.

lassik: I don't think there is any unique identifier one can retrieve for a modern computer that's likely to be both accurate and privacy-respecting.

Emacs tries (or tried) to infer the user's email address based on the hostname and username and that often failed. Peronal computers don't tend to have a hostname that makes sense.

Some device ID (e.g. Ethernet MAC address) would probably be the most reliable in practice, but poking around hardware details like that is behavior that is nowadays mainly attributed to malware.

src/ldeboot.c
```diff
@@ -200,12 +200,19 @@ int main(int argc, char *argv[]) {
...
   } else {
     /* copy up to and including the final "/" in the path */
-#ifndef MAIKO_OS_HAIKU 
-    dirsepp = stpncpy(filetorunpath, argv[0], dirsepp + 1 - argv[0]);
-#else
-    dirsepp = strncpy(filetorunpath, argv[0], dirsepp + 1 - argv[0]);
+
+#ifdef MAIKO_OS_HAIKU
+    /* Copied from libiberty stpncpy.c */
+    char * stpncpy (char *dst, const char *src, size_t len) {
```
nbriggs, line 206: Umm...  function definition is not allowed here.  What C compiler are you using??

nbriggs: Am i correct in assuming that the original problem in this area was that Haiku doesn't provide a definition for `stpncpy()` in libroot?  If it's present, but in another library, link against that.  If it isn't present at all, also file a bug report against either Haiku itself (to get stpncpy implemented), or the libroot documentation that suggests it implements POSIX requirements but doesn't, and doesn't provide a list of unimplemented functions (stpncpy is a POSIX function).

src/initsout.c
lassik: How about making it a variable whose default value is either zero or a random number?

lassik: People who care about it could assign a value of their choice. Perhaps in a config file.

masinter: There is code in multiplayer LOOPS that I remember seeing that relies on stable hostid across savevm and restore (but not sysout or makesys presumably)

nbriggs: @masinter - hmm.  I see truckin/LOOPSGATEWAY setting up `(SETQ \MY.NSHOSTID (CONCAT "#NSID." (CADR \MY.NSHOSTNUMBER)` but I don't see any direct reference to the InterfacePage serial number, or to the Unix getenv/getparm (which would be unexpected since it wouldn't rely on there being a Unix system underneath, right?).

src/ldeboot.c
waddlesplash: Indeed it appears Haiku is missing `stpncpy`. However, it's an "Issue 7" (2018) addition, and Haiku's `_POSIX_VERSION` is `200809L`, meaning we don't make any guarantees about compliance past then. (We have implemented a lot of POSIX-2018 but not all of it yet.)

I will work on getting `stpncpy` added to Haiku's `libroot`.

waddlesplash: >  What C compiler are you using??

Inline function definition like this is a GCC extension, IIRC.

src/timer.c
```diff
@@ -268,6 +272,13 @@ static int gettime(int casep)
 /*	Implements the SETTIME subr call, sub-dispatched from subr.c	*/
 /*									*/
 /************************************************************************/
+#ifdef MAIKO_OS_HAIKU
+int settimeofday(struct timeval *tv, struct timezone *tz)
+{
+  set_real_time_clock(tv->tv_sec);
```
waddlesplash, line 278: If `timezone` is ignored, it would probably be best for this method to be a no-op; just have it return `-1` and set `errno` to `EOPNOTSUPP`.

src/initsout.c
waddlesplash: Reproducing the full comment from the Haiku bug tracker:

> - The implementation in glibc attempts to compute an hostid based on the local IP, but on Debian it computes it based on 127.0.0.1 instead so all machines have the same host id.
> - musl just returns 0
> - NetBSD, FreeBSD and OpenBSD implements it but mentions that is deprecated and that it had been removed in 4.4BSD 

Note that some Linux distributions (e.g. Alpine Linux) use `musl` as the default C library. And of course Debian is not an unpopular distribution...

masinter: @nbriggs I didn't remember the details, I just flagged this as a possible reason why Truckin didn't find player names... that unique IDs weren't necessarily unique?
https://github.com/Interlisp/medley/issues/260 has more details.

src/ldeboot.c
lassik: https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

src/timer.c
nbriggs: @waddlesplash it only ever passes `NULL` to the `settimeofday()` call for the timezone.  The call-site doesn't care about the success/failure or any return value - it's a best-effort and isn't really expected to work on Unix-y systems, it really only mattered when Lisp *was* the operating system and then it could do the right thing (and that wasn't via settimeofday syscalls!)

src/ldeboot.c
nbriggs: Yeah, ok.  We try to stick to ISO/ANSI C.  I ported the code forward from K&R C to C89, and then we've started to use C99 in some cases, but ISO C forbids nested functions.  Some GCC extensions have crept in (e.g., computed goto in the `-DOPDISP` case) but we try to keep them at an absolute minimum to preserve portability/compatibility with older compilers and systems.

nbriggs: @waddlesplash `stpncpy()` is present in The Open Group Base Specifications Issue 7, IEEE Std 1003.1-2008, which says _POSIX_C_SOURCE has value 200809L, so I'm surprised it's not there.

waddlesplash: @nbriggs I can't find it in the 2008 edition. Do you have a link? I can only find it in the 2018.

src/timer.c
waddlesplash: Well, if it isn't really expected to work, this really should be reduced to a no-op. `set_real_time_clock` is certainly not a no-op on Haiku!

src/ldeboot.c
nbriggs: @waddlesplash - https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/ - but search seems to shift to a more recent edition, so here's the direct link - https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/stpncpy.html

But thanks for pressing on with adding it to the Haiku library.

bin/makefile-haiku.x86_64-sdl
```diff
@@ -0,0 +1,19 @@
+# Options for Linux, Intel x86_64 and SDL
```
jmairboeck, line 1: Haiku isn't Linux.

bin/makefile-haiku.x86_64-x
```diff
@@ -0,0 +1,28 @@
+# Options for Linux, Intel x86_64 and X-Window
```
jmairboeck, line 1: the same here

src/ldeboot.c
Sylvain78: Haiku for stpncpy in hrev57422

src/timer.c
```diff
@@ -275,7 +275,8 @@ static int gettime(int casep)
 #ifdef MAIKO_OS_HAIKU
```
nbriggs, line 275: OK, but there's exactly ONE call to `settimeofday`, from `subr_settime()` also in timer.c, where it already has `#ifdef DOS ... #else...` to handle systems that have differing implementations for the time-of-day setting.  You can just eliminate all code except a `return;` in the Haiku case.  There's no checking of the return value - if it works, it works, otherwise nothing in the code cares.

src/initsout.c
Sylvain78: As per [https://dev.haiku-os.org/ticket/10841](https://dev.haiku-os.org/ticket/10841), Haiku won't implement gethostid(), so i advise to stick as the same behaviour as DOS.

nbriggs (approved):
This looks fine.

PR #486, opened 2023-12-16
XFree(mapping) was called *before* mapping was referenced …

MattHeffron:
…in DEBUG conditional code.

This is only a potential issue in DEBUG build. A problem hasn't occurred, that I know of, but it seemed best just to be safe. Moved the DEBUG code to precede the XFree(mapping) call. (Equivalent to moving the XFree down.)


nbriggs (approved):
Thanks.  Looks fine.

PR #488, opened 2024-01-25
Incorporate wasm changes

nbriggs:
Integrates the WASM work done by others more fully into the system for building and packaging.

In order to build (from maiko/bin, ./makeright wasm) you need the emscripten package(s) installed, with a working emcc, which will also require a version of clang (this setup is system specific and not documented here).  The "makeright" will generate, in the same style as other targets, the emscripten.wasm directory, containing ldesdl.js, ldesdl.wasm, and ldesdl.data.  The HTML page that wraps these is (now) stored in maiko/src/lde.html.   In order to run lde under wasm you need a web server in which you will install all 4 files.  As it is currently set up, all the files go at the root of the web directory.  If your web server is on a host named "lisp.local", you would then browse http://lisp.local/lde.html to run Medley.  When the lde files for WASM are built they capture the state of the adjacent `medley` directory -- all the files in docs, doctools, greetfiles, internal, sources, library, lispusers, fonts, as well as loadups/full.sysout,  loadups/whereis.hash, and greetfiles/MEDLEYDIR-INIT.LCOM -- so you need to make sure you have built a native maiko, and done the medley loadup to get a full.sysout, before you build the WASM lde.

I'd appreciate it if someone who has the capability of testing this would do so...


masinter (approved):
LGTM

image

I was able to do a clean install of wsl and emscript etc so...
LGTM

PR #490, opened 2024-02-12
Emscripten/wasm does not support getrusage() system call

nbriggs:
Updating the MiscStats data can at best provide the time of day and attribute all elapsed time to total CPU time.

Closes issue #1537


fghalasz (approved):
LGTM.  Tried it out online.  Console msgs gone and functionality seems unchanged.

waywardmonkeys (approved):
As an addendum, this isn't something that can be picked up via a configuration time check since emscripten does implement the function and everything at the level which is externally visible. (Since this is really part of `musl` which then makes a syscall which emscripten just stubs out.)

PR #491, opened 2024-02-16
Fix Issue#1541: IL:LOGOUT does not work running Medley on emscripten maiko

fghalasz:
The function vmem_save uses the construction `if ( errno != 2) ...` (in 2 places).

On most systems, errno 2 is ENOENT.  But for some reason on emscripten, ENOENT is errno 44.

Changed construction to `if ( errno != ENOENT) ...` for all platforms (i.e., not conditionalized on MAIKO_OS_EMSCRIPTEN).  Shouldn't really have hard-wired errnos in cross-platform code.


nbriggs (approved):
LGTM.  This prompted me to check for any other errno comparisons to numbers other than 0 and fortunately there are none.

PR #492, opened 2024-02-16
Update makefile-tail and makefile-emscripten.wasm-wasm to fix emscripten build support for IDBFS

fghalasz:
1. Fixed makefile-tail so it uses `$(LD)` for linker operations instead of `$(CC)`
   `$(LD)` is set to `$(CC)` if `$(LD)` not set before makefile-tail.
   Allows separate linker and cc commands.  Emscripten makefile is set up like this,
   although its not strictly necessary.
 
2. In makefile-tail added `$(UPFRONT_LDFLAGS)` to accomodate emscripten build
   where -lidbfs.js needs to come before files in ld call.

3) Replaced wasm builds with wasm_nl builds; removed wasm_nl builds as per suggestion a couple of PRs back.


fghalasz:
Closing this PR for now.  Open prematurely.  Too many problems still to be worked out.

bin/makefile-emscripten.wasm-wasm
```diff
@@ -1,16 +1,16 @@
...
+CC = emcc $(CLANG_CFLAGS)
 
 XFILES = $(OBJECTDIR)sdl.o
 
 XFLAGS = -DSDL -sUSE_SDL=2
 
 # OPTFLAGS is normally -O2.
 OPTFLAGS = -O2
-DFLAGS = $(XFLAGS) -DRELEASE=351 -DMAIKO_ENABLE_NETHUB
+DFLAGS = $(XFLAGS) -DRELEASE=351 -DMAIKO_OS_EMSCRIPTEN -DMAIKO_ENABLE_NETHUB
```
nbriggs, line 11: By the way: If you get back to this, no makefile should define any `MAIKO_OS_...` flag, these should be determined automatically in `inc/maiko/platform.h`.  In this case, `emcc` always sets both `__EMSCRIPTEN__` and `EMSCRIPTEN`, which triggers `platform.h` to set up the various necessary defines.  The `MAIKO_ENABLE_...` are optional features that are determined externally, typically not as a requirement of the compilation environment, though some may be supplied as with `MAIKO_ENABLE_NETHUB` here.

fghalasz: Thx.  I discovered platform.h when trying to update this PR.  But in the end, I've decided to just abandon this branch and start over on adding the IDBFS libraries to the emscripten builds.  Most of what I did on this branch was just unnecessary - the right solution is MUCH simpler.  I was just tripped up by caching in the browser while trying out various build options.

nbriggs: Ah, yes, caching caught me a couple of times too.  I took to deleting all the files from the nginx directory and recreating them so that nginx would pass on the new creation date and the browser wouldn't use the cached versions.

waywardmonkeys: One should only be checking for `__EMSCRIPtEN__` and not the version without underscores.

nbriggs: Yup, that is the check in maiko/platform.h --
```
#ifdef __EMSCRIPTEN__
#  define MAIKO_OS_LINUX 1
#  define MAIKO_OS_EMSCRIPTEN 1
#  define MAIKO_OS_NAME "Emscripten"
...
#endif
```

PR #493, opened 2024-03-16
Cleanup of stack-related code

nbriggs:
Adds one additional validation in `StackOffsetFromNative()` -- a stack offset can be bad if it is negative, not just too large.

Cleans up adding/oring of a constant to do conversion from LispPTR to stack offset and replaces these operations with calls to a function (inlined) that encapsulates the knowledge of the conversion.


masinter (approved):
reviewed 3/18/24 meeting

PR #497, opened 2024-03-28
Update SDL dependent code and makefile segments to prepare for SDL3

nbriggs:
Many APIs have changed between SDL2 and SDL3.  This update adds preprocessor conditionals to adapt the Maiko SDL code to allow selection of the SDL major version (2 or 3) from the
`-DSDL=n` define.  The expectation is that when SDL 3 is released and has become widely available the processor conditionals could be removed.

The SDL3 implementation is currently available as a preview release, 3.1.0, at
   [SDL3 prerelease 3.1.0](https://github.com/libsdl-org/SDL/releases/tag/prerelease-3.1.0)


nbriggs:
Needs CMake update to support SDL 2/3 choice.  Marked as draft until that is completed.

nbriggs:
See [CMake issue 25816](https://gitlab.kitware.com/cmake/cmake/-/issues/25816)

nbriggs:
NOTE that installing the SDL3 framework for macOS requires installing both the `SDL3.xcframework` AND `share` folders into `/Library/Frameworks` in order for CMake to work, though there is still an issue with CMake for which we have a workaround (noted in the CMakeLists.txt).

PR #498, opened 2024-04-02
Improve the file lookup performance for {DSK} files

nbriggs:
The directory caching for file lookup on {DSK} was disabled due to a bug in SunOS - this is no longer relevant, so the cache can be reenabled.


nbriggs:
@masinter, @rmkaplan - you might find that this maiko branch speeds up loadups, in particular for the loadup-aux part of the process, plus anything else where a lot of files in a `{DSK}` based directory are accessed without new files being written.

PR #500, opened 2024-05-07
Update config.guess, config.sub to current GNU versions

nbriggs:
The versions of config.guess and config.sub we were shipping were from 2020.  Since then the scripts have been cleaned up, made mostly POSIX compliant (pass shellcheck with some noted exceptions), and recognize some newer systems.

These scripts should be backward compatible, but please test if you have any unusual systems.

### DIFF
--- a/bin/config.guess
+++ b/bin/config.guess
@@ -1,12 +1,14 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2020 Free Software Foundation, Inc.
+#   Copyright 1992-2024 Free Software Foundation, Inc.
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys (approved):

nbriggs:
@fghalasz - could you check these on your windows/cygwin configurations please?

fghalasz:
@nbriggs config.guess, osversion and machinetype work just fine on cygwin/x86_64.

(Have no way of testing and have never run Medley on cygwin/aarch64.)

Also, not sure who calls/uses config.sub, so wasn't really able to test.

nbriggs:
@fghalasz thanks.

PR #501, opened 2024-05-07
IPv4 host network addresses convert with htonl() not htons()

nbriggs:
In the host based TCP subrs, converting a FIXP to an IPv4 address was never going to work calling htons(), since addresses are 32-bit not 16-bit.  Use htonl().  This looks like a day-0 bug.

### DIFF
--- a/src/inet.c
+++ b/src/inet.c
@@ -26,6 +26,7 @@
 #include <sys/time.h>
 #include <sys/types.h>
 #include <unistd.h>
+/* if using inet_ntop you must #include <arpa/inet.h> */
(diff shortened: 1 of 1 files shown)


nbriggs:
Don't merge this until the previous PR, #500  has been merged

waywardmonkeys (approved):

waywardmonkeys:
Curious why this depends on #500.

nbriggs:
@waywardmonkeys - you're right, they are independent.  I forgot that I *had* switched back to master before creating the 2nd one.

PR #502, opened 2024-05-08
Not all systems define MAP_FAILED for the return value from failing mmap()

nbriggs:
If `MAP_FAILED` is not defined after including `<sys/mman.h>` then define it to be `(void *)-1`


waywardmonkeys:
What system isn't defining this? It seems to be SuSv2, so over 25 years old at this point...

nbriggs:
AmigaOS

waywardmonkeys:
Probably deserves a comment or something since that is pretty obscure.

PR #503, opened 2024-05-10
Rationalize the processing of command line args for the sysout.  Fixes Issue#1703

fghalasz:
Fixes Issue#1703 and more.

Rationalize the processing of command line args for the sysout.  Remove most of the redundancy between what is done for sysout arg processing in xrdopt.c versus main.c.  Move all of the decision making about what sysout_name to use until all args have been processed (both in xrdpopt and main).  This is done in main, where the prioritization and checking is done in one place rather than spread in several places in xrdopt and main.

 Order of priority is specifying sysout to lde::
1. Value of -sysout command line arg
2. Value of the first command line arg
3. Value of LDESRCESYSOUT env variable
 4. Value of LDESOURCESYSOUT env variable
 5. Value as determined by X resource manager, if any
6.  Value of $HOME/lisp.virtualmem (or lisp.vm for DOS)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -229,7 +229,11 @@ int display_max = 65536 * 16 * 2;
 /* diagnostic flag for sysout dumping */
 extern unsigned maxpages;
 
-char sysout_name[MAXPATHLEN]; /* Set by read_Xoption, in the X version. */
+char sysout_name_cl[MAXPATHLEN] = "\0"; /* sysout name as per -sysout command line arg ; Set by read_Xoption, in the X version. */
+char sysout_name_xrm[MAXPATHLEN] = "\0"; /* sysout name as per X resource manager, if any */
+char sysout_name_first_arg[MAXPATHLEN] = "\0"; /* sysout name as per 1st command line arg, if any */
+char sysout_name[MAXPATHLEN] = "\0"; /* "final" sysout name chosen from above */
+
 unsigned sysout_size = 0;    /* ditto */
 
 int flushing = FALSE; /* see dbprint.h if set, all debug/trace printing will call fflush(stdout) after each printf */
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -369,34 +367,23 @@ int main(int argc, char *argv[])
...
-    perror("Couldn't find a sysout to run");
-    (void)fprintf(stderr, "%s%s", helpstring, nethubHelpstring);
-    exit(1);
-  }
-  /* OK, sysout name is now in sysout_name, and i is moved past a supplied name */
+
 
   for (; i < argc; i += 1) { /* step by 1 in case of typo */
 
+   /* Check for -sysout arg */
```
nbriggs, line 377: This could have a comment noting that it will not happen in the X11 case, but is handling -sysout for SDL or any other display subsystem that doesn't parse its own command line for -sysout.

src/main.c
```diff
@@ -610,6 +597,40 @@ int main(int argc, char *argv[])
     }
   }
 
+  //
+  //  OK, now we can process the sysout_name
+  //  Order of priority:
+  //    1. Value of -sysout command line arg
+  //    2. Value of the first command line arg
```
nbriggs, line 604: Though because of the way the X11 options parser works, it's actually the first non-X11 accepted argument.  This will work perfectly:
```
lde -title "foo" lisp.sysout -geometry 500x500+20+20
```

src/main.c
fghalasz: That is true for many (most?) of the arguments being processed in this argument processing loop - they have already been processed and removed from argv by the call to read_Xoption.  They are "repeated" in this argument processing loop for the non-X platforms.  I added a comment at the top of the loop to note that this is the case.

PR #504, opened 2024-05-10
Fix Cygwin github builds:  Issue#1705

fghalasz:
In buildRelease.yml, change site for cygwin download to https:/mirrors.kernel.org since old gatech site is now deprecated.


fghalasz:
Merging now since needed to fix cygwin medley installs.

PR #505, opened 2024-05-10
Add -rpath /Library/Frameworks to accommodate newer macOS releases

nbriggs:
On or before macOS Ventura's release, defaulting the @rpath to include /Library/Frameworks stopped.  Since we expect the SDL2 framework to be installed there we must add a -rpath option. This should be backwards compatible with older macOS releases. (verified on Catalina but nothing older)


nbriggs:
Fixes issue Interlisp/medley#1706

nbriggs:
Tested on macOS monterey.  No problem.

PR #506, opened 2024-05-20
Implement the "B hexaddr" command to print array block

nbriggs:
The array block printer was already documented, but not implemented.
Clean up with IncludeWhatYouUse, to document why particular includes are present.


masinter (approved):
as per 5/20/24

PR #507, opened 2024-05-20
Update .gitignore to ignore 386-sdl build directory

nbriggs:
Trivial update to add the directories created by builds for SDL based display on 386 class systems.


masinter (approved):
as per 5/20/2024

PR #508, opened 2024-05-22
Add support for setting cursor color to ldex (X11 display)

nbriggs:
The cursor (foreground) color can be set with the
    -cursorColor x11-color-spec
option to ldex or any front-end that passes the options through to ldex. For example:
     lde ... -cursorColor red
     lde ... -cursorColor '#A020F0'
     ./run-medley -cursorColor RED -g 1200x700 -sc 1200x700 -noscroll

Alternatively, the color may be specified with the ldex*cursorColor resource (in ~/.Xresources, or other such X11 resource databases that are loaded by the X11 server)

The command line option will override the X11 resource option.

### DIFF
--- a/src/xcursor.c
+++ b/src/xcursor.c
@@ -26,6 +26,7 @@
 
 static XColor cursor_fore_xcsd, cursor_back_xcsd, xced;
 extern Colormap Colors;
+extern char cursorColor[255];
 
 extern DspInterface currentdsp;
 /* a simple linked list to remember X cursors */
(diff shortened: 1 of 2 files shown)


hjellinek (approved):
I didn't try to build it, but it looks sane and solid.
+1

masinter:
Worked for me although setting it up as a default needed some script fiddling... it doesn't follow the Linux -- vs - convention @fghalasz does it?

PR #509, opened 2024-08-16
Ignore mouse motion events when the window does not have keyboard focus

nbriggs:
This addresses a problem wherein mouse motion in a Medley SDL window that does not have the keyboard focus slows down drastically if the cursor moves over a TEdit window in the exposed area.

The change is to detect the SDL keyboard focus events and stop reporting mouse motion events to Medley when it does not have the keyboard focus.

Closes Interlisp/medley#1807


hjellinek (approved):
Read and tested and approved!

PR #510, opened 2024-08-29
Release 201 corrections

nbriggs:
Corrections to code and improvements to build infrastructure to enable building older releases of Maiko compatible with older Medley sysouts.

* `makeright` will use the value of the environment variable `RELEASE` to select the version to build (default current, 351)
* `CMake` option `MAIKO_RELEASE` selects the version to build if using that tool (default current, 351)
* corrects a parameter type in `modify_big_reference_count` to reflect that the size of a GC table entry depended on the version.

The build artifacts and directories do NOT include the release version number, so be careful to clean before building.  An enhancement here could be to add the version for non-current build directories and executable names.

While the code now compiles, attempting to run an older (2.0) sysout does not work, so more investigation and fixes remain to be done.


pamoroso:
It would be useful to have some notes on the new build options, for example by copying [this PR's description](https://github.com/Interlisp/maiko/pull/510#issue-2495375204) to the Maiko README.

nbriggs:
A complete update of the README.md would be good.  There's no SDL, Networking, or CMake info there either.

However, building for an old release is not (yet) interesting for anyone who isn't pretty familiar with the history of Medley releases, the various changes that the different versions implied (big atoms, bigvm, bigbigvm, etc.) and helping debug the Maiko code.

masinter (approved):
is there value in merging this without the additional documentation. I suppose so.

PR #511, opened 2024-09-03
Allow choice of foreground and background colors for Medley X11/SDL display

nbriggs:
Accepts -fg/-foreground <color> and -bg/-background <color>, or the corresponding X resources ldex\*foreground and ldex\*background to specify the colors that will represent the monochrome Medley display pixels.  The same color syntax, \#RRBBGG (hex) or a named color in the X11-style, will work with the command line arguments for the SDL display subsystem case.


nbriggs:
This needs to be extended to the SDL display subsystem.  It's a bit harder there because SDL does not provide a named color API nor is there built-in option and resource parsing.

PR #512, opened 2024-09-11
Fix #1826 - emulator crash  when too many versions of a file in a directory

nbriggs:
This is a preliminary fix to avoid a system crash (bus error/seg fault) when operating on a file with too many versions present in the directory.

Reports an error (EIO, SIMPLE-DEVICE-ERROR) when doing an operation that would result in an out-of-bounds access in the emulator.

A better fix would be to dynamically allocate the necessary version storage, but that will
take longer to work out.

Test environment is a directory that contains over 200 versions of a file and you attempt to do a create (new version) or delete operation on the file.

### DIFF
--- a/src/dsk.c
+++ b/src/dsk.c
@@ -2941,7 +2941,7 @@ static int make_directory(char *dir)
 static int get_version_array(char *dir, char *file, FileName varray[], CurrentVArray *cache)
 {
 #ifdef DOS
-
+  /* DOS version-array builder */
   char lcased_file[MAXPATHLEN];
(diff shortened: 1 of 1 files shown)


nbriggs (commented):
@MattHeffron - I read through the DOS part again, and I believe it is correct.
At the only point where it increments the `varray_index` value it checks that it remains within bounds, and the last value that it takes on (and no greater) *will* be used to access the array at that index when the loop exits.  If the value goes out of range - well, that's where I have no idea how to report the value upstream.

PR #513, opened 2024-10-01
Cleanup in the "version array" handling in the {DSK} device

nbriggs:
Removed the unused caching of version array entries but then reimplemented it using nanosecond resolution directory modification timestamps which will avoid the issue of an outdated cache failing to be detected.

Simplified setting of file modification times on close through use of the `futimens()` call which operates on the open file descriptor directly rather than having to find the file name after closing.

Switch to chunked allocation (200 entries) for the version array, maintaining an upper limit on the number of versions of a file (2000 entries).  There's no particular basis for these limits other than things will get pretty slow if you're working with 2000 versions of a file.  The numbers can be adjusted by changing `VERSIONARRAYCHUNKLENGTH` and `VERSIONARRAYMAXLENGTH`.

Sort (quicksort) the version array, in decending version order, when it is created.  `FindHighestVersion` is a frequent operation and it can make use of the ordering.

### DIFF
--- a/inc/locfile.h
+++ b/inc/locfile.h
@@ -499,7 +499,8 @@ do  {				\
 #define	MAXVERSION		999999999
 
 #define	LASTVERSIONARRAY	((unsigned) -1)
-#define	VERSIONARRAYLENGTH	200
+#define	VERSIONARRAYCHUNKLENGTH	200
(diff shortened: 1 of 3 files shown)


nbriggs:
There is more cleanup to be done here, but I'd like to get this tested (and merged) first.

PR #519, opened 2024-12-14
Fix display bitmap flushing on modification when using SDL display subsystem

nbriggs:
Fixes issue Interlisp/medley#1908

In cases where the target of a `PILOTBITBLT` opcode, or the target of an `FBITMAPBIT` opcode, is the display bitmap, the SDL display subsystem needs to be notified of the "damage" so that the real display can be updated.

Slightly Improves performance for `FBITMAPBIT` in case where there is no soft cursor that would need to be taken down before accessing the screen bitmap.

### DIFF
--- a/src/bitblt.c
+++ b/src/bitblt.c
@@ -126,7 +126,7 @@ LispPTR N_OP_pilotbitblt(LispPTR pilot_bt_tbl, LispPTR tos)
   if (displayflg) (currentdsp->mouse_visible)(IOPage->dlmousex, IOPage->dlmousey);
 #endif /* SUNDISPLAY / DOS */
 
-#ifdef XWINDOW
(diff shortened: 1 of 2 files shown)


hjellinek:
I just pulled the branch, built maiko SDL, and resumed the virtual memory image I had used to create the test case.  Now `CIRCLE-BUG` displays the circles immediately.  I then built maiko for X and tested that way with identical results.

Nice!  Thanks for the quick fix, @nbriggs.

@abhikhasnain1, once this is merged you should get a new Maiko and run with that.  Let us know if you need help with that.

nbriggs:
@hjellinek do you want to "officially" review/approve before I merge it, or should I just take the comment above as approval.

hjellinek (approved):
J'approuve. It fixes the reported problem.

PR #520, opened 2024-12-17
Replaces shadowed "-t <interval>" with "-timer <interval>"

nbriggs:
At some point option "-t" was introduced as an abbreviation for "-title" but "-t" was already in use, though undocumented, for overriding the timer interval.  This commit changes the command line timer option to "-timer" and adds "timer" to the options that are processed from the X11 resources.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -407,20 +407,20 @@ int main(int argc, char *argv[])
       }
     }
 
-    /* -t and -m are undocumented and somewhat dangerous... */
+    /* -timer and -m are undocumented and somewhat dangerous... */
 
-    else if (!strcmp(argv[i], "-t")) { /**** timer interval	****/
+    else if (!strcmp(argv[i], "-timer")) { /**** timer interval	****/
       if (argc > ++i) {
         errno = 0;
         tmpint = strtol(argv[i], (char **)NULL, 10);
         if (errno == 0 && tmpint > 0) {
           TIMER_INTERVAL = tmpint;
         } else {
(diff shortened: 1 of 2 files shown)


masinter:
the handling of lde call parameters vs. the ones to the scripts/medley parameters confuses me.
Should anyone be still using the "run-medley" shell script?   Maybe allow --timer as well as -timer if you want consistency?

nbriggs:
I use run-medley.  I don't use "medley.sh" (?).  The scripts to drive lde started at PARC and continued as they have here, for similar reasons -- people wanted shortcuts so it mostly did the right thing without having to think about it -- but we've been changing the script options to spell them differently (e.g., double-dash-long-option instead of dash-long-option and dash-short-option, with new names) because people don't like the old option names (or didn't know what old options there were so they invented new ones).

The lde options provide all the knobs for the low level things that the higher level scripts may make coordinated changes to (e.g., "-id foo" sets up the window title as "foo", and sets up the LDESRCSYSOUT/LDEDESTSYSOUT environment variables to "~/lisp/lisp-foo.virtualmem", and so on...)

We could make a breaking change and switch to all "--foo" options on lde/ldex/ldesdl/ldeether, but I'm not exactly sure what that buys us, except for either breaking the X11 standard option parsing, which has builtin options (like "-geometry") or having mixed use of dash and double-dash options.

So... for this case, I just made the expedient fix that meant that if we wanted to experiment with the timer frequency, we didn't have to edit the source and recompile maiko because the only timer option was shadowed by the short form title option.

PR #524, opened 2025-01-15
Modifies Lisp exit subr to permit passing an exit status code

nbriggs:
These changes add `(LOGOUT <number>) `to the traditional (LOGOUT) and (LOGOUT T) calls.  If an integer `<number>` is provided it will be used as the exit status code, while NIL and T result in an EXIT_SUCCESS.

If the argument passed is none of NIL, T, or a number, the exit status code will be a generic EXIT_FAILURE (typically 1).

(LOGOUT) and (LOGOUT T) virtual memory behavior is unaffected. When a `<number>` is passed Lisp will exit without saving the virtual memory state, as is the case for (LOGOUT T), although this behavior is determined by Lisp rather than the Maiko emulator.

### DIFF
--- a/inc/uutilsdefs.h
+++ b/inc/uutilsdefs.h
@@ -6,5 +6,5 @@ LispPTR unix_username(LispPTR *args);
 LispPTR unix_getparm(LispPTR *args);
 LispPTR unix_getenv(LispPTR *args);
 LispPTR unix_fullname(LispPTR *args);
-LispPTR suspend_lisp(LispPTR *args);
+LispPTR suspend_lisp(void);
 #endif
--- a/inc/vmemsavedefs.h
+++ b/inc/vmemsavedefs.h
@@ -4,5 +4,5 @@
 int lispstringP(LispPTR Lisp);
 LispPTR vmem_save(char *sysout_file_name);
 LispPTR vmem_save0(LispPTR *args);
-void lisp_finish(void);
+void lisp_finish(int exit_status);
 #endif
--- a/src/subr.c
+++ b/src/subr.c
@@ -29,6 +29,7 @@
 /***********************************************************/
 
 #include <stdio.h>         // for printf, sprintf, NULL
+#include <stdlib.h>        // for EXIT_FAILURE, EXIT_SUCCESS
 #include <time.h>          // for nanosleep, timespec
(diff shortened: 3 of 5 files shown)


fghalasz:
Adding comment late.  Sorry.

Might make sense to separate the two arguments (Save/NoSave and Exit Code) in the API.  The function signature would be (LOGOUT NoSave ExitCode).   Default for NoSave in NIL and for ExitCode is 0.  This way (LOGOUT) and (LOGOUT T) would maintain their current semantics.  And you could still save vm even with an exit code  (whether or not that exit code is 0).

nbriggs:
@fghalasz - That should be fairly easy, edits to `LOGOUT` and `\LOGOUT0` in ADIR should do it.  The current implementation doesn't change the semantics of (LOGOUT) and (LOGOUT T), you always get an exit code of 0, as you did previously -- and (LOGOUT 0) is the same as (LOGOUT T) (because 0 is not NIL, the Lisp code won't SAVEVM).  I was trying to avoid changing the Lisp code, but your suggestion makes it more usable.

nbriggs:
@fghalasz - I can do this with compatibility between old/new Maiko and old/new sysouts --  In all cases the savevm behavior of the fast flag (NIL/T) is maintained, but note the old undocumented old behavior of fast = 0 that I'll remove, there's a separate (SUSPEND-LISP) subr for that.

New maiko:
old sysout:
   * (LOGOUT), (LOGOUT T) will exit with status 0 (EXIT_SUCCESS, as currently) because they'll (SUBRCALL LISPFINISH FAST) (arg count = 1 to SUBR)
   * (LOGOUT fast exitcode) in old sysout, same, since exitcode doesn't get passed to the SUBRCALL (arg count = 1)

new sysout:
   * (LOGOUT fast) will exit with status 0,  (arg count = 2) exit code NIL => 0
   * (LOGOUT fast exitcode) exitcode is a number, will exit with status `exitcode`
   * (LOGOUT fast exitcode) exitcode is NOT a number, will exit with status EXIT_FAILURE (1).
 
For an older maiko (that knows nothing about this scheme)
    either an old or new sysout:

   * (LOGOUT fast) for fast NIL or T exits with status code 0, 
   * (LOGOUT 0) does a suspend rather than an exit -- this is old behavior that is undocumented
   * (LOGOUT fast exitcode) ignores exitcode, behaves as above.

PR #526, opened 2025-01-26
Correct {DSK} versioned file cache failures

nbriggs:
A missing check for the applicability of a versioned file cache caused loadups to fail in the case where the modification timestamp of a directory and its parent were identical, and the file being looked up had the same name (when lowercased) as the directory it was contained within. E.g., looking up TEDIT in library/tedit/ where the modification times of library and library/tedit were identical.

The inode number of the directory containing the file must be the same as the inode number of the directory the cache was built from.

### DIFF
--- a/src/dsk.c
+++ b/src/dsk.c
@@ -66,6 +66,7 @@ typedef struct filename_entry {
  */
 static struct {
   char name[MAXPATHLEN];     /* lowercase unversioned file name */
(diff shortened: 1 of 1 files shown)


fghalasz (approved):
Ok.  Works as expected - TEdit loads fine during all loadups and loadup-all.sh -apps completes successfully.

PR #527, opened 2025-01-29
Fix MacOS SDL build in buildRelease github workflow

fghalasz:
Updated the Macos SDL build in buildRelease workflow to use `-DMAIKO_DISPLAY_SDL=2` instead of `DMAIKO_DISPLAY_SDL=ON`.

Also updated SDL Framework used for build to version 2.30.11 (from 2.26.5).


nbriggs (approved):
Looks good.  Thanks for the fix.

PR #532, opened 2025-02-17
Enable runtime controlled debug output.

MattHeffron:
Currently only in `make_X_keymap`.
Output is to `STDOUT` (since the conditional compilation `DEBUG` did)

Medley branch: **mth38--command-line-arg-to-support-maiko-runtime-debug** adds command line argument support for this.

This is draft as it probably isn't generally applicable. OTOH, it's also relatively harmless (as of now).

### DIFF
--- a/src/initkbd.c
+++ b/src/initkbd.c
@@ -71,6 +71,8 @@ extern DspInterface currentdsp;
 extern int LispKbdFd;
 int LispKbdFd = -1;
 
+extern int runtime_debug_level;
+
 extern fd_set LispReadFds;
 
 extern DLword *EmMouseX68K;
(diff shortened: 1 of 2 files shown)


nbriggs (commented):
Since this doesn't depend on anything other than the results from XGetKeyboardMapping (e.g., it doesn't reflect maiko's interpretation of its own table with respect to this), I think having this in a trivial extra program would be better - in the same way we ship programs like "tstsout.c" and "setsout.c"

MattHeffron:
That's fine with me for the current output from `initkbd.c`. 
If this is a useful _concept/feature_ for maiko, should the changes in `main.c` be kept? (A separate PR?)

Specifically, being able to turn on some _pre-defined_ debugging output without needing to rebuild.
Maybe also, do either bit-testing or exact value testing, _vs._ > threshold testing against the `runtime_debug_level`. (That would mean that it isn't a `_level`.)

If there's no value in this at all, then there's no need for Medley PR [#2038](https://github.com/Interlisp/medley/pull/2038)

nbriggs:
My choice would be something akin to (if not actually) DTrace probes.  Before DTrace existed we used a very low-overhead always-on debugging probe mechanism which maintained a circular buffer of events which you could dump either in real-time, or extract from a core file.  Anything that goes into the dispatch loop for the emulator needs to be carefully considered, and the probe mechanism should introduce the minimum possible latency.

MattHeffron:
Close/delete this PR (and Medley PR 2038) as unnecessary (at least at this time).

PR #535, opened 2025-05-28
Fix Issue#2164 - mouse does not work on WSL1.  Root cause is that WSL1 does not support setitimer(ITIMER_VIRTUAL ...).

fghalasz:
Fix Issue#2164 - mouse does not work on WSL1.  Root cause is that, like cygwin, WSL1 does not support setitimer(ITIMER_VIRTRUAL ...)…  

WSL1 compiles using gcc and clang as straight linux and WSL1 cannot be detected using any preprocessor macros.  Also don't want to create yet another platform (WSL1) to run through the whole github actions release process. 

So in the case of MAIKO_OS_LINUX (including WSL1), we now detect setitimer failure and use emulated timer when such failure occurs. So unlike CYGWIN and Emscripten, the decision to use emulated timers for WSL1 is made at runtime not compile time.  This will add a very small overhead to all non-WSL1 Linux runs but should not effect any other platforms.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -315,7 +315,7 @@ const char *nethubHelpstring =
 const char *nethubHelpstring = "";
 #endif
 
-#if defined(MAIKO_EMULATE_TIMER_INTERRUPTS) || defined(MAIKO_EMULATE_ASYNC_INTERRUPTS)
+#if MAIKO_OS_LINUX || defined(MAIKO_EMULATE_TIMER_INTERRUPTS) || defined(MAIKO_EMULATE_ASYNC_INTERRUPTS)
 extern int insnsCountdownForTimerAsyncEmulation;
 #endif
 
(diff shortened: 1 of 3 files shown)


nbriggs:
n.b. - WSL2 was released 6 years ago.

fghalasz:
@nbriggs  Compiling with -DMAIKO_EMULATE_TIMER_INTERRUPTS -DDMAIKO_EMULATE_ASYNC_INTERRUPTS is not straight-forward.  It involves editing the right bin/makefile-linux-... file before running makeright since there is no way to pass these macro defs thru makeright.

Alternative is to create a makefile-wsl1.x86_64-x with these two macros defined and then modify osversion to detect and return "wsl1".   Then, at least, there would be a straightforward way to compile and run maiko on WSL1.

But I am not sure what the issue is with the runtime detection that I've added.  Its restricted to Linux platforms only, which keeps things mostly as is but makes the ifdefs a bit more complicated than necessary.  The ifdef-ness of the code could be cleaner overall if we did runtime detection for all platforms including Cygwin and Emscripten.  If an initial call to setitimer(ITIMER_VIRTUAL ...) fails then set things to run with an emulated timer.  The cost, if not timer emulation, would be one simple 'if gate' each time around the dispatch loop.

fghalasz:
Closing this PR, unmerged.   Moving to a compile time solution to handling WSL1's lack of a virtual interval timer instead of the run-time solution used in the PR.  This compile time solution will be released as a separate PR in short order.

PR #536, opened 2025-06-04
To fix Issue#2164, create infrastructure to compile Maiko for Windows System for Linux v1 as a separate OS from Linux

fghalasz:
WSL1 does not support a virtual interval timer (similar to Cygwin).  As a result, the standard linux build for Maiko results is a non-functional Maiko on WSL1 - as documented in Issue #2164.

This PR creates separate builds for WSL1, which are identical to the Linux builds except for the use of the simulated interval timer currently used in Cygwin builds to replace the missing virtual interval timer.

Changes are as follows:
1.  Modified bin/osversion to return "wsl1" when running on WSL1 and "linux" on all other linux platforms. 
2.  Added settings for the WSL1 platform to inc/maiko/platform.h.
3. Created the various makefile-wsl.* files to support compiling on WSL1 using makeright.  (Cmake not yet supported.)
4. Modified the buildRelease.yml workflow and its attendant Dockerfile_maiko so that when a release is built, it will include separate WSL1 builds for x86_64 and aarch64.

Changes to support the new WSL1 platform in Medley releases will be included in an upcoming PR is the Medley repo.
Also PR#2180 in Medley repo, removes the forced use of vnc on WSL1 - handy for testing this PR.

### DIFF
--- a/.github/workflows/Dockerfile_maiko
+++ b/.github/workflows/Dockerfile_maiko
@@ -29,7 +29,6 @@ RUN cd ${INSTALL_LOCATION}/bin \
            export LDEARCH="${arch}-microsoft-wsl1" \
            && ./makeright x cleanup \
            && ./makeright x \
-           && ./makeright init clean \
            && ./makeright init \
            ; \
        fi


masinter:
@nbriggs please review and marge if approve. Some medley PRs depend on this one.

PR #539, opened 2025-07-21
Minor cleanup of code style in gcfinal.c and additional array free list debugging from URaid

nbriggs:
This set of changes adds a new URaid command, `F [size]`, which will display the address of the first block on the free list for the bucket holding the given size, or, if no size is specified, all 31 buckets.  For example,
```
< F
[...]
Free block chain (bucket 3): NIL
Free block chain(bucket 4): 0x3cffe2
Free block chain (bucket 5): NIL
Free block chain(bucket 6): 0x3fffbe
Free block chain(bucket 7): 0x40aeb6
Free block chain (bucket 8): NIL
Free block chain(bucket 9): 0x45061c
Free block chain (bucket 10): NIL
[...]
< F 500
Free block chain(bucket 9): 0x45061c
```
Using this, one can walk the free block chain using the (existing) `B address` command which dumps an array block header and some words around the beginning and end of the block:
```
< B 0x45061c
This array block: 0x45061c.  Previous: 0x450450.  Next: 0x450956.
        password: 0x1555     gctype: 0x0   in use: 0
      Free list: fwd 0x45061c bkwd 0x45061c
  Header  Length: 413 cells.

 Trailer  Length: 413 cells.
[...]
```
In addition to this added functionality, the changes
   * introduce a more consistent naming, in some of the GC code, of variables that represent the native pointer equivalent of a Lisp address (e.g., `base`/`base_np`, `trailer`/`trailer_np`).  This convention could be used generally through the code in the future.
   * improve the readability of some nested code in GC routines by rewriting `if (test) .../else return;` forms as `if (!test) return; ...` (the "early out" pattern)

### DIFF
--- a/inc/gcfinaldefs.h
+++ b/inc/gcfinaldefs.h
@@ -5,7 +5,6 @@ void printarrayblock(LispPTR base);
 LispPTR releasingvmempage(LispPTR ptr);
 LispPTR checkarrayblock(LispPTR base, LispPTR free, LispPTR onfreelist);
 LispPTR makefreearrayblock(LispPTR block, DLword length);
-LispPTR arrayblockmerger(LispPTR base, LispPTR nbase);
 LispPTR mergebackward(LispPTR base);
 LispPTR mergeforward(LispPTR base);
 LispPTR reclaimarrayblock(LispPTR ptr);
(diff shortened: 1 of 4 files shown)


nbriggs:
That's annoying.  "fls" has been in BSD for over 20 years, and Linux C libraries have not picked it up?

nbriggs:
@pamoroso should be fixed now.

pamoroso:
I confirm commit 01d21b0 fixed. Now Maiko builds and runs fine.

PR #540, opened 2025-08-05
string handling cleanup

nbriggs:
Start at cleaning up some of the string handling in Maiko.  So far this only eliminates one routine (StrNCpyFromLispToC).

There should be no change in the normal operation of the code, the only new feature is that passing an indirect string (such as generated by a displaced array substring on a string) to URaid will print it correctly as the generic object printer will decode indirect strings.


pamoroso:
I successfully built the code of this branch on Linux Mint 22.1 Cinnamon and Medley is working fine.

PR #544, opened 2026-03-18
Consolidate PRs 542 and 543 and slightly update:  Redo linux makefiles to include libbsd where necessary and to abstract out common parts and share them across all linux-y makefiles

fghalasz:
This PR consolidates PRs 542 and 543 (which have been closed).  The PR incorporates the following:

1.  All linux and wsl1 Makefiles now (when glibc < 2.38) compile and link with libbsd using the libbsd-overlay scheme via pkg-config.  This is to enable the use of strlcat and friends, which are not included in glibc prior to v2.38.

2. All Linux, cygwin and  wsl1 makefiles have been refactored so that common code is factored out of the top level (e.g., makefile-linux.x86_64-x) makefiles and moved to several common code makefiles (linux-*.mk) that are included into the top level makefiles.

4.  All linux, cygwin and wsl1 makefiles will now automatically select between gcc and clang at runtime using the following rule:  if call to make includes USE_COMPILER=gcc or USE_COMPILER=clang then use that compiler otherwise use clang, if it is installled, and gcc if clang is not installed.  (Error, of course, if neither installed.)

4.   Cmakelists now has special code for Linux that (when glibc < 2.38) causes compile and link to use the libbsd-overlay scheme via pkg-config.

5.  Added an install target to CMakeLists that installs lde, ldeinit ldex and ldesdl in the same binary directories that the Makefiles do, e.g., maiko/linux.x86_64.

6.  Update all github workflows to use the latest versions of github actions to account for the impending deprecation of Node 20 on github.

7.  Made all changes to 542 and 543 that were suggested by Gemini Code Check.

8.  Fixed issue with $(shell command -v pkg-config) on Ubuntu.


nbriggs (changes requested):
Still has the `command` problem (it doesn't work on the system that online.interlisp.org is running) - the problem occurs in linux-compiler.mk and linux-libbsd.mk.

nbriggs (approved):
LGTM now.

